PR #18, opened 2023-05-11
feat(HACBS-2107): Add Network Policy to internl-service controller

happybhati:
The following commit adds an Egress Network Policy to the internal-service app-interface cluster,
only allowing traffic to listed `ipBlock` and blocking the rest of the traffic from the cluster for better security

### DIFF
--- a/internal-services/config/networkpolicy/production/network_policy.yaml
+++ b/internal-services/config/networkpolicy/production/network_policy.yaml
@@ -1,0 +1,38 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: egress-external-cluster
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: controller-manager
+  policyTypes:
+  - Egress
+  egress:
+    - to:
+      - namespaceSelector: {}
+        podSelector:
+          matchLabels:
+            apiserver: "true"
+            app: openshift-kube-apiserver
+      ports:
+        - port: 6443
+        - port: 80
+        - port: 443
+          protocol: TCP
+    - to:
+      - ipBlock:
+          cidr: 44.195.126.208/32  # Hypershift-Dev
+      ports:
+        - port: 6443
+        - port: 80
+        - port: 443
+          protocol: TCP
+    - to:
+      - ipBlock:
+          cidr: 54.91.12.133/32  # RHTAP-Prod
+      ports:
+        - port: 6443
+        - port: 80
+        - port: 443
+          protocol: TCP
--- a/internal-services/config/networkpolicy/staging/network_policy.yaml
+++ b/internal-services/config/networkpolicy/staging/network_policy.yaml
@@ -1,0 +1,38 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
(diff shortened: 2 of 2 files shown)


internal-services/manager/network_policy.yaml
```diff
@@ -0,0 +1,46 @@
...
+            apiserver: "true"
+            app: openshift-kube-apiserver
+      ports:
+        - port: 6443
+        - port: 80
+        - port: 443
+          protocol: TCP
+    - to:
+      - ipBlock:
+          cidr: 44.195.126.208/32  # Hypershift-Dev
```
scoheb, line 25: will this have to change when using a dev environment each time?

happybhati: - Yes, we can either make a shared release-dev cluster to avoid the change to policy, 
- or we need to  perform `nslookup <cluster API address(without https)>`
Example: `nslookup abcd.elb.us-east-1.amazonaws.com`
on the cluster which is configured in the vault, and add the `IPAddress` from `nslookup` to `cidr` ipBlock of network policy.

scoheb: I'd prefer option 2 as long as we have a comment on the network policy yaml.

I would also add a comment to the ISC config Google doc about this.

scoheb (changes requested):
See comment

internal-services/manager/network_policy.yaml
happybhati: I have added the Network Policy config section in the [Doc](https://docs.google.com/document/d/1ElMVAgHEZ0NHcy9QAJtJAd6LrSY_n8Al18oMPE3STm0/edit#) as requested.

scoheb (approved):

internal-services/config/networkpolicy/staging/network_policy.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - ipBlock:
+          cidr: 52.202.122.112/32  # RHTAP-Stage
+      ports:
+        - port: 6443
+        - port: 80
+        - port: 443
+          protocol: TCP
+    - to:
+      - ipBlock:
+          cidr: 54.91.12.133/32  # RHTAP-Prod
```
scoheb, line 41: remove and only leave staging

internal-services/config/networkpolicy/production/network_policy.yaml
```diff
@@ -0,0 +1,46 @@
...
+            apiserver: "true"
+            app: openshift-kube-apiserver
+      ports:
+        - port: 6443
+        - port: 80
+        - port: 443
+          protocol: TCP
+    - to:
+      - ipBlock:
+          cidr: 44.195.126.208/32  # Hypershift-Dev
```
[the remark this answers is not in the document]
happybhati: @scoheb 
Do we need to keep this(Hypershift-Dev) for testing? Once we merge the change and if we need to test if we can create internalRequest to prod/stage we need to keep dev at least until we test.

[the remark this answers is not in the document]
scoheb: ok let's keep for now

scoheb (approved):

PR #19, opened 2023-05-18
feat(HACBS-2205): add new fromImage credentials

scoheb:
Refactor secrets to extract fromImage token from config secret and have it isolated.


happybhati (approved):
LGTM 👍

PR #24, opened 2023-05-22
fix: remove jq -cr param to keep result as string

theflockers:
Fix the output saved in `buildState` result to be always a regular string. For some reason the `-cr` param is making openshift understand the result as an object.

https://console-openshift-console.apps.appsres03ue1.5nvu.p1.openshiftapps.com/k8s/ns/stonesoup-int-srvc/tekton.dev~v1beta1~PipelineRun/internalrequest-cph6h

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -156,7 +156,7 @@ spec:
             esac
         done
         echo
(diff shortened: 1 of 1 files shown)


scoheb (dismissed):
This just seems to be a revert of a change I added to try to fix problem?

theflockers:
> This just seems to be a revert of a change I added to try to fix problem?

I confess I did not notice this reverts  a change you made. I will crosscheck if anything else is required to fix this.

scoheb (approved):

PR #29, opened 2023-05-31
feat(HACBS-2207): Production IIB pipeline only overwrites fromImage and copies to targetImage when opted-in

theflockers:
HACBS-2207: Production IIB pipeline only overwrites fromImage and copies to
targetImage when opted-in.

- changes task `t-add-fbc-fragment-to-index-image` to check if FBC index
  is optIn instead of using `overwriteFromIndex` parameter
	- removes `overwriteFromIndex` parameter
	- updates `release-base-image` version
	- adds `misc` result
- adds `publish-index-image` pipeline
- adds `task-publish-index-image` task

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -19,9 +19,6 @@ spec:
       type: string
       description: ->
         Index image (catalog of catalogs) the FBC fragment will be added to
-    - name: overwriteFromIndex
-      type: string
-      description: Boolean indicating if the from_index should be overwritten
     - name: binaryImage
       type: string
       description: ->
@@ -53,10 +50,12 @@ spec:
       description: JSON build information for the requested build
     - name: buildState
       description: IIB Service build state
+    - name: genericResult
+      description: Set the genericResult if FBC Fragment is Opt-In and should be published
   steps:
     - name: s-add-fbc-fragment-to-index-image
       image: >-
-        quay.io/hacbs-release/release-base-image@sha256:9e7fd1a3ccf0d2c8077f565c78e50862a7cc4792d548b5c01c8b09077e6d23a7
+        quay.io/hacbs-release/release-base-image@sha256:5968660e005b0c806dd2568b6c8b488b4b519ed33daa41211b9f9f2bde4faf35
       env:
         - name: IIB_SERVICE_URL
           valueFrom:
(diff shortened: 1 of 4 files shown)


internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -81,6 +80,37 @@ spec:
...
+
+          # strips the last "/tag" in case $TAG is not set
+          [ -z "${TAG}" ] && FETCH_URL=${FETCH_URL%/tag*}
+
+          curl --insecure -v --negotiate -u: "${FETCH_URL}" -o $TMPFILE
+          # prints "false" in case .fbc_opt_in entry is missing
+          jq -e -r '.fbc_opt_in //false' $TMPFILE && rm -f $TMPFILE
+        }
+
+        PULL_SPEC=$(opm render $(params.fbcFragment) | jq -r 'select(.schema == "olm.bundle") | "\(.image)" | split("@")[0]' |uniq)
```
scoheb, line 100: you do not handle the case where there might be 2 elements in your query

scoheb (changes requested):

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: I did not get what you mean @scoheb . Do you mean if the `opm` return a list of images with different hashes? This query does Just like your example but not using `tr` (because it is ugly :D). 

If it is not that, can you elaborate  what I actually did not cover?

theflockers: Or you mean, even after reducing the result I still have more than one image? In this case it sounds like a problem. Would it be possible to have different `opt_in` value for each of the images?

scoheb: It is possible that you may get 2 images after being reduced. This can happen if the image got renamed and/or moved in quay.

If you get 2, then both need to be opt_in.

theflockers: Understood. So it means, for any reason one of them are not opt_in, should the Pipeline fail or continue without publishing the index?

theflockers: Please take a look again @scoheb

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -53,10 +50,12 @@ spec:
       description: JSON build information for the requested build
     - name: buildState
       description: IIB Service build state
+    - name: misc
```
scoheb, line 53: why `misc`? what does that mean?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -93,26 +134,18 @@ spec:
...
         
         # filtering out empty params
         jq -r '
           if .binary_image == "" then del(.binary_image) else . end |
-          if .overwrite_from_index == "false" then del(( .overwrite_from_index, .overwrite_from_index_token)) else . end |
+          if .overwrite_from_index == false then del(( .overwrite_from_index, .overwrite_from_index_token)) else . end |
           if(.add_arches | length) == 0 then del(.add_arches) else . end |
           if(.build_tags | length) == 0 then del(.build_tags) else . end' ${json_raw_input} > ${json_input}
         
-        # performs kerberos authentication.
```
scoheb, line 108: not needed?

internal-services/catalog/iib-pipeline.yaml
```diff
@@ -83,3 +76,5 @@ spec:
       value: $(tasks.t-add-fbc-fragment-to-index-image.results.jsonBuildInfo)
     - name: buildState
       value: $(tasks.t-add-fbc-fragment-to-index-image.results.buildState)
+    - name: misc
```
scoheb, line 79: `misc` again?

scoheb (changes requested):

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: I had to add a result to transfer the opt in information to the internalrequest. In the `create-internal-request` it is called also called`misc` therefore I kept the same name.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -93,26 +134,18 @@ spec:
           "binary_image": "$(params.binaryImage)",
           "build_tags": `echo $(params.buildTags)`,
           "add_arches": `echo $(params.addArches)`,
-          "overwrite_from_index": $(params.overwriteFromIndex),
+          "overwrite_from_index": ${FBCOptIn},
           "overwrite_from_index_token": "${IIB_OVERWRITE_FROM_INDEX_USERNAME}:${IIB_OVERWRITE_FROM_INDEX_TOKEN}"
```
scoheb, line 138: does IIB complain if you give a token but overwrite is `false`?

theflockers: Yes. It does. So the next part of the script has a json that checks if the value is false and remove both parameters.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -81,6 +80,37 @@ spec:
...
         #!/usr/bin/env bash
         #
+        isFBCOptIn() {
+          TMPFILE=$(mktemp)
+          PYXIS_URL="https://pyxis.engineering.redhat.com/v1"
+
+          IFS="/" read REGISTRY REPO IMAGE <<< "${1}"
+          IFS=":" read IMAGE TAG <<< "${IMAGE}"
+
+          FETCH_URL="${PYXIS_URL}/repositories/registry/${REGISTRY}/repository/${REPO}/${IMAGE}/tag/${TAG}"
```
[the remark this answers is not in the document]
scoheb: I see the new log messages! thanks!

PR #30, opened 2023-06-21
feat(HACBS-2226): create a task to clone a repo and update files given fileUpdates for repo.

theflockers:
feat(HACBS-2226): create a task to clone a repo and update files given fileUpdates for repo.

### DIFF
--- a/internal-services/catalog/file-updates-process-file-updates-task.yaml
+++ b/internal-services/catalog/file-updates-process-file-updates-task.yaml
@@ -1,0 +1,126 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: task-process-file-updates
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: release
+spec:
+  description: >-
+      Update files in a Git repository
+  params:
+      - name: repo
+        type: string
+        description: Git repository
+      - name: ref
+        type: string
+        description: Git branch
+      - name: paths
+        type: string
+        description: |
+            String containing a JSON array of file paths and its replacements
+            E.g. '[{"path":"file1.yaml","replacements":[{"key":".yamlkey1,","replacement":"|regex|replace|"}]}]'
+  results:
+      - name: fileUpdatesInfo
+        description: fileUpdates detailed information
+      - name: fileUpdatesState
+        description: fileUpdates state
+  steps:
+      - name: perform-updates
+        image: quay.io/hacbs-release/release-utils@sha256:17bf3669322362e3fef010bfb4f2b636a56e89d15880e027ca16eba3e6f9d16f
+        env:
+          - name: GITLAB_HOST
+            valueFrom:
+                secretKeyRef:
+                    name: file-updates-secret
+                    key: gitlab_host
+          - name: ACCESS_TOKEN
+            valueFrom:
+                secretKeyRef:
+                    name: file-updates-secret
+                    key: gitlab_access_token
+          - name: GIT_AUTHOR_NAME
+            valueFrom:
+                secretKeyRef:
+                    name: file-updates-secret
+                    key: git_author_name
+          - name: GIT_AUTHOR_EMAIL
+            valueFrom:
+                secretKeyRef:
+                    name: file-updates-secret
+                    key: git_author_email
+        script: |
+            #!/usr/bin/env sh
+            set -eo pipefail
+
+            # loading git and gitlab functions
+            . /home/utils/gitlab-functions
+            . /home/utils/git-functions
+
+            gitlab_init
+
+            # saves the params.paths json to a file
+            updatePathsTmpfile="/tmp/updatePaths.json"
+            cat > "${updatePathsTmpfile}" << JSON
+            $(params.paths)
+            JSON
+
+            REPO="$(params.repo)"
+            REVISION="$(params.ref)"
+
+            echo -e "=== UPDATING ${REPO} ON BRANCH ${REVISION} ===\n"
+
+            cd /tmp/
+            git_clone_and_checkout --repository "${REPO}" --revision "${REVISION}"
+
+            # getting the files that have replacements
+            PATHS_LENGTH="$(jq '. | length' ${updatePathsTmpfile})"
+            for (( PATH_INDEX=0; PATH_INDEX < $PATHS_LENGTH; PATH_INDEX++ )); do
+                targetFile="$(jq -cr ".[${PATH_INDEX}].path" ${updatePathsTmpfile})"
+
+                # we need to know how many empty newlines and `---` the file has before the actual yaml data starts excluding comments
+                blankLinesBeforeYaml="$(awk '/[[:alpha:]]+/{ if(! match($0, "^#")) { print NR-1; exit } }' ${targetFile})"
+
+                # getting the replacements for the file
+                REPLACEMENTS_LENGTH="$(jq -cr ".[${PATH_INDEX}].replacements | length" ${updatePathsTmpfile})"
+                for (( REPLACEMENT_INDEX=0; REPLACEMENT_INDEX < $REPLACEMENTS_LENGTH; REPLACEMENT_INDEX++ )); do
+                    key="$(jq -cr ".[${PATH_INDEX}].replacements[${REPLACEMENT_INDEX}].key" ${updatePathsTmpfile})"
+                    replacement="$(jq -cr ".[${PATH_INDEX}].replacements[${REPLACEMENT_INDEX}].replacement" ${updatePathsTmpfile})"
+
+                    # getting the key's position
+                    echo -en "Searching for key \`${key}\`: "
+                    foundAt="$(yq "${key} | line" ${targetFile})"
+                    if (( ${foundAt} == 0 )); then
+                        echo "NOT FOUND"
+                        continue
+                    fi
+                    echo "FOUND"
+
+                    # getting the value size (in number of lines)
+                    valueSize="$(yq "${key}" ${targetFile} |wc -l)"
+                    startBlock=$(( ${foundAt} + ${blankLinesBeforeYaml} ))
+
+                    # run the replace
+                    sed -i "${startBlock},+${valueSize}s${replacement}" "${targetFile}"
+                done
+            done
+
+            echo -e "\n*** START LOCAL CHANGES ***\n"
+            git diff
+            echo -e "\n*** END LOCAL CHANGES ***\n"
+
+            WORKING_BRANCH=`uuidgen |awk '{print substr($1, 1, 8)}'`
+            git_commit_and_push --branch $WORKING_BRANCH --message "fileUpdates changes"
+
+            echo "Creating Pull Request..."
+            GITLAB_MR_MSG="fileUpdates changes ${WORKING_BRANCH}"
+            gitlab_create_mr --head $REVISION --title "chore: ${GITLAB_MR_MSG}" \
+                --description "${GITLAB_MR_MSG}" > $(results.fileUpdatesInfo.path)
+            echo "Success" |tee $(results.fileUpdatesState.path)
+
+            echo -e "=== FINISHED. CLEANING UP ===\n"
+            cd /tmp
+            rm -rf ${OLDPWD}
--- a/internal-services/catalog/file-updates-pipeline.yaml
+++ b/internal-services/catalog/file-updates-pipeline.yaml
@@ -1,0 +1,41 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: file-updates
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: release
+spec:
+  description: >-
+      Tekton Pipeline to update files in Git repositories
+  params:
+    - name: repo
+      type: string
+      description: Git repository
+    - name: ref
+      type: string
+      description: Git branch
+    - name: paths
+      type: string
+      description: |
+          String containing a JSON array of file paths and its replacements
+          E.g. '[{"path":"file1.yaml","replacements":[{"key":".yamlkey1,","replacement":"|regex|replace|"}]}]'
+  tasks:
+    - name: task-process-file-updates
+      taskRef:
+        name: task-process-file-updates
+      params:
+        - name: repo
+          value: $(params.repo)
+        - name: ref
+          value: $(params.ref)
+        - name: paths
+          value: $(params.paths)
+  results:
+    - name: jsonBuildInfo
+      value: $(tasks.task-process-file-updates.results.fileUpdatesInfo)
+    - name: buildState
+      value: $(tasks.task-process-file-updates.results.fileUpdatesState)


mmalina:
@theflockers the story only talks about creating a task, not a pipeline. So maybe the pipeline is not needed at this stage? Also, it says that the format should be clearly defined. I know that you basically took what I had in my story (and I took it from the google doc linked from the feature/epic). But it would be good if this gets documented somewhere for future reference. Not sure about the best place though. Maybe it should be here? https://redhat-appstudio.github.io/book/book/release-service.html

theflockers:
@mmalina I agree that the ticket does not say it, although to test it  I had to create the Pipeline as the InternalServices controller can create the InternalRequest to run it. That said, I would rather rewrite the ticket adding the Pipeline to it. Maybe @scoheb can tell us what he thinks.

About the doc, agreed as well. Not sure about the best place.

The functions will be in another PR. I will link it here once I push it.

scoheb:
we'll eventually add the tasks to https://github.com/redhat-appstudio/release-service-bundles/tree/main/catalog/pipeline/push-to-external-registry

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,95 @@
...
+      - name: fileUpdatesInfo
+        description: fileUpdates detailed information
+      - name: fileUpdatesState
+        description: fileUpdates state
+  workspaces:
+      - name: temp
+        description: Temporary working directory
+  steps:
+      - name: perform-updates
+        image: quay.io/lmendes/base-image@sha256:205ccc0f254894ffb69f0ed6dd19aa35805953d4a26eedce5c3234a696cfcd2a
```
theflockers, line 28: Just adding a note here. This will be changed once the`release-base-image` PR is merged.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -72,14 +72,21 @@ spec:
...
-                   REPLACEMENT=$(jq -cr '[.yamlpath, .replace] | join(" ")' <<< $replacement)
-                   read YAMLPATH REPLACE <<< $REPLACEMENT
-                   LN=`python3 /home/utils/yamlline.py --file $file --path $YAMLPATH`
-                   LINES=$(echo -e `yaml-get -p "$YAMLPATH" $file` | wc -l)
+                   IFS=$'\t' read KEY REPLACE <<< $replacement
+                   # the key position
+                   LN=`yq "${KEY} | line" $file`
+
+                   # the number of lines + overhead to run the replace 
+                   LINES=$(expr `yq "${KEY}" $file |wc -l` + ${OVERHEAD})
```
mmalina, line 87: Am I reading this right that you don't care if the empty lines are before our field or after? You will just add as many extra lines to the range as the number of empty lines you find, just to be sure to cover the section you want? But that also means that you may include some lines you don't want, right? Which could be an issue - the sed could change something other than what you want.

theflockers: Indeed. I will change to only add the blank newlines until the position. I thought about that yesterday but forgot to make the changes. Thanks for saying it!

theflockers: Noticed now that actually `yq` is giving the position but with just 1 line less for all cases. I tested  several different keys  in the bottom of the that deploy file and still just 1 line of difference. Please take a look.

mmalina: I don't think that's the case. For example I have a file like this:
```yaml







resourceTemplates:
 - name: osd-fleet-manager-production
 ```

And I get:

```sh
$ cat test.yaml | yq '.resourceTemplates[0] | line'
2
```

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -73,24 +73,29 @@ spec:
...
-                   LN=`yq "${KEY} | line" $file`
-                   if [ "${LN}" ==  0 ]; then
+                   foundAt=`yq "${KEY} | line" $file`
+                   if [ "${foundAt}" -eq 0 ]; then
                        echo "NOT FOUND"
                        continue
                    fi
                    echo "FOUND"
 
                    # yq always gives the pos -1 line, so we add it
```
theflockers, line 93: this comment makes no sense anymore, I will remove.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,117 @@
...
+            set -eo pipefail
+
+            # loading git and gitlab functions
+            . /home/utils/gitlab-functions
+            . /home/utils/git-functions
+
+            gitlab_init
+
+            # saves the fileUpdates raw json to a file
+            TEMP=$(mktemp).json
```
mmalina, line 61: I checked `mktemp` documentation and this is not how it should be used. mktemp will generate a unique filename and also create it for you to use. You can provide a template to use, e.g. `mktemp /tmp/myjson.XXXX` and it will replace the `XXXX` with some unique/random characters. But here, given that each step starts with a fresh container, it's fine to just use a file in the current dir. (I'd also use a better name for the variable.) Like:

```sh
UPDATESFILE=fileUpdates.json
```

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - name: fileUpdates
+        type: string
+        description: fileUpdates json string
+  results:
+      - name: fileUpdatesInfo
+        description: fileUpdates detailed information
+      - name: fileUpdatesState
+        description: fileUpdates state
+  workspaces:
+      - name: temp
```
mmalina, line 24: What's the workspace for? Usually it's used when you need to share data between steps or even tasks. But it doesn't seem to be happening here - you're just cloning the repos there, but they could be cloned anyway. Or is it in preparation for a new step that will create the MRs? In that case it would be ok, but I would propose something less generic for the name, e.g. `repos`.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+            gitlab_init
+
+            # saves the fileUpdates raw json to a file
+            TEMP=$(mktemp).json
+            cat > $TEMP <<JSON
+            $(params.fileUpdates)
+            JSON
+
+            REPOSITORY=$(jq -cr '[.repo, .ref] | join(" ")' $TEMP)
```
mmalina, line 66: Wait, so the `fileUpdates` is not the whole `fileUpdates` section from the CR, it's just a subset for one repo? Then the description should make it clear.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,117 @@
...
+            read REPO REVISION <<< $REPOSITORY
```
mmalina, line 67: I would suggest simply loading the two variables directly - it's also two lines, but more straightforward.
```sh
REPO=$(jq -cr '.repo' $TEMP)
REVISION=$(jq -cr '.ref' $TEMP)
```

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+            FILES=$(jq -cr '.paths[] | .path' $TEMP)
+
+            echo -e "=== UPDATING ${REPO} ON BRANCH ${REVISION} ===\n"
+            cd $(workspaces.temp.path)
+            git_clone_and_checkout --repository $REPO --revision $REVISION
+
+            for file in ${FILES}; do
```
mmalina, line 75: It's not great that we dig up the paths from inside the `paths` array and then in the for cycle we are looking up the array items again. It could also cause issues if a single path has several items (which shouldn't be the case, but it could happen).

Since this is an array, I would suggest using indices to access the items in the for cycle.
```sh
PATHS_LENGTH=$(jq '.paths | length' $TEMP)
for PATH_INDEX in seq 0 $((PATHS_LENGTH-1)); do
```

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,117 @@
...
+                # in case the file has comments, empty newlines and other special chages in the
+                # start of the file we need to know where the actual yaml data starts
+                blankLines=`awk '/[[:alpha:]]+/{ if(! match($0, "^#")) { print NR-1; exit } }' $file`
+
+                jq -cr ".paths[] | select(.path==\"${file}\") | .replacements[] | [.key, .replacement] | join(\"\t\")" $TEMP| \
+                while read replacement; do
```
mmalina, line 81: Similarly to iterating over `paths`, I would suggest to iterate over `replacements` indices.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+            # loading git and gitlab functions
+            . /home/utils/gitlab-functions
+            . /home/utils/git-functions
+
+            gitlab_init
+
+            # saves the fileUpdates raw json to a file
+            TEMP=$(mktemp).json
+            cat > $TEMP <<JSON
```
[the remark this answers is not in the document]
theflockers: This way bash does not have the change to remove or add quotes or escape chars where is should not. But I will give a try again.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: the input of this will indeed be different after our discussion yesterday in the Stand Up. David made the adjusts required in the `internal-request` script and I will implement this as it should. Working on it atm.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -73,24 +73,29 @@ spec:
...
                        echo "NOT FOUND"
                        continue
                    fi
                    echo "FOUND"
 
-                   # yq always gives the pos -1 line, so we add it
-                   LINES=$(expr `yq "${KEY}" $file |wc -l` + 1)
+                   # getting the value size (in number of lines)
+                   valueSize=$(expr `yq "${KEY}" $file |wc -l`)
+                   startBlock=`expr $foundAt + $blankLines`
```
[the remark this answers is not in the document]
theflockers: I will move to use`$()` for all cases.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -73,24 +73,29 @@ spec:
...
+                   if [ "${foundAt}" -eq 0 ]; then
                        echo "NOT FOUND"
                        continue
                    fi
                    echo "FOUND"
 
-                   # yq always gives the pos -1 line, so we add it
-                   LINES=$(expr `yq "${KEY}" $file |wc -l` + 1)
+                   # getting the value size (in number of lines)
+                   valueSize=$(expr `yq "${KEY}" $file |wc -l`)
```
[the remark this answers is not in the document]
theflockers: Indeed. Forgot to remove it.

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
mmalina: I don't think bash will add or remove anything when you run `echo "$(params.fileUpdates)"`. 

Or this is another option:

```sh
cat > $TEMP <<< "$(params.fileUpdates)"
```

I tried it with env vars and it worked fine even if there were double quotes in the strings (e.g. a json string). But here it might be different, because here the param substitution is done by tekton and you might end up with something like:
```sh
echo "{"a": "b"}"
```
which wouldn't work. So it needs testing.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: Also, not sure if it ok to rely in the space available in the container to fetch the repo. It might be huge depending on the case.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -73,24 +73,29 @@ spec:
             git_clone_and_checkout --repository $REPO --revision $REVISION
 
             for file in ${FILES}; do
+                # in case the file has comments, empty newlines and other special chages in the
```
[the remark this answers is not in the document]
theflockers: That's right. The thing is that the  yq's `|line`  function  removes all whitespaces until the first valid json, but keeps comments and `---`. After it, it preserves all of them. Notice that if you do `yq . <file>`  the output does not match this behavior as it removes only comments *after* the first valid json. At least this is the behavior in `v.4.34.1` which is the one in the release-utils image.

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
theflockers: I tried  `echo "$(params.paths)" > $TEMP` and it broke exactly in the echo. I guess this is not a problem with the bash itself, but because this `$(params.xxx)` is a tekton "macro".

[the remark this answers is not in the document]
theflockers: ` 68             cat > $TEMP <<< "$(params.paths)"` also broke the same way. It is now `params.paths`  because I am not using the full `fileUpdates` as param anymore.

The error FYI

```
+ TEMP=/tmp/tmp.m3yFeDDpF3.json
/tekton/scripts/script-0-d9cph: line 15: syntax error near unexpected token `('
```

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: changed.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: changed.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: changed.

theflockers:
PTAL @mmalina

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
mmalina: OK, thanks for testing it :)

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: This can be changed now, right? Since no further changes in the image will be required.

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
mmalina: Thanks, it seems this was done.

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
mmalina: ok, cool

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: > Also, not sure if it ok to rely in the space available in the container to fetch the repo. It might be huge depending on the case.

How big is the workspace now? Do you know? The PipelineRun is created by an operator and the operator provides a PV, right?

I was trying to find out what is the default size of the ephemeral storage provided for the Pods in Tekton, but couldn't find it here: https://tekton.dev/docs/pipelines/compute-resources/ It would be nice if we could compare those.

But this brings another point - I think we should definitely use shallow clones here: https://github.com/redhat-appstudio/release-service-utils/blob/main/utils/git-functions#L71 That will save a lot of space. Can you create a PR for that?

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: ok, cool. Can you maybe provide more details on the expected format of `paths` in the parameter description? It could even include a sample snippet.

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: Thanks. But looking at the current version, it seems like the variables can be removed completely? I don't think having them adds much value. WDYT?

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: Much better, thanks.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: For now I changed to /tmp. I agree with your point on the shallow clone, but If you don't mind, I prefer to open a ticket and make the changes when I am back in two weeks, so I can at least close this ticket which is there for about a month.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: it is already the proper image.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: This is fixed.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: I would keep just to make it easier in case of upcoming changes that might change variable names or these tekton variables are needed to be used in more places.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: sure. I will do.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: I might do it still today, but just in case we take to long to close this.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: done.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: https://github.com/redhat-appstudio/release-service-utils/pull/51

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: ok, cool

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: yep, thanks

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: ok, fine

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: thanks

internal-services/catalog/file-updates-process-file-updates-task.yaml
mmalina: We cleared this up.

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
mmalina: Hmm, I'm pretty sure I added a comment here a few hours ago. Maybe I forgot to hit the Comment button? Weird.

> Notice that if you do yq . <file> the output does not match this behavior as it removes only comments after the first valid json.

I don't think it necessarily removes comments. See this snippet (it's the output of `yq . test.yaml` and the comment on the last line is preserved):

```yaml
---






resourceTemplates:
  - name: osd-fleet-manager-production
    parameters:
      CPU: [8, 10]
      MEMORY_LIMIT: 8Gi
    targets:
      - name: osd-fleet-manager-production-us-east-2
        # this is my comment
```

What I still don't understand is what this is for: `if(! match($0, "^#"))`, or I mean the whole `awk` command. I read it as a comment that starts at the beginning of the line. But it seems the `#` is actually a special character in the regex? Like `$` elsewhere? But I have a hard time finding it in the docs.

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -0,0 +1,126 @@
...
+                targetFile="$(jq -cr ".[${PATH_INDEX}].path" ${updatePathsTmpfile})"
+
+                # we need to know how many empty newlines and `---` the file has before the actual yaml data starts excluding comments
+                blankLinesBeforeYaml="$(awk '/[[:alpha:]]+/{ if(! match($0, "^#")) { print NR-1; exit } }' ${targetFile})"
+
+                # getting the replacements for the file
+                REPLACEMENTS_LENGTH="$(jq -cr ".[${PATH_INDEX}].replacements | length" ${updatePathsTmpfile})"
+                for (( REPLACEMENT_INDEX=0; REPLACEMENT_INDEX < $REPLACEMENTS_LENGTH; REPLACEMENT_INDEX++ )); do
+                    key="$(jq -cr ".[${PATH_INDEX}].replacements[${REPLACEMENT_INDEX}].key" ${updatePathsTmpfile})"
+                    replacement="$(jq -cr ".[${PATH_INDEX}].replacements[${REPLACEMENT_INDEX}].replacement" ${updatePathsTmpfile})"
```
mmalina, line 91: I wonder if we should at least somehow validate that the `replacement` is a valid string for `sed` to consume. Maybe we can add something later.

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
theflockers: Let-me clarify it in another way. 
Forget about the comments. The idea of this is: I need to know where the actual yaml block starts because `yq |line` will consider that spot as line 1. 
The "#" part is because we might have alphanumeric chars there, for instance`# blabla` which is not a valid start of yaml as  I noticed. Another case is that you might have `$schema` which is a valid word for start the yaml.

Once I know that is the start, I just get the position found -1 (which is the line above). So, from the start of the file, until there, all was ignored by `yq |line`.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: I agree. Also on doing this later. I can add a ticket for this.

internal-services/catalog/file-updates-process-file-updates-task.yaml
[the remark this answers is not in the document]
mmalina: We discussed this later in a chat. It turns out that in some cases the behavior of `yq` can be inconsistent. If you add indented comments before the first key in the yaml, sometimes `yq` will count the line and sometimes not. But either way such comment is probably invalid yaml, because the indentation should be always aligned with some key/element.

Anyway, let's go with this for now and we can improve it later.

mmalina (approved):

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: ticket create and added to the backlog.

PR #37, opened 2023-07-07
fix: adjusts to fit the new pre-ga release strategy

theflockers:
FBC pre-GA release strategy requires different credentials, so we should be able to set them.

### DIFF
--- a/internal-services/catalog/iib-pipeline.yaml
+++ b/internal-services/catalog/iib-pipeline.yaml
@@ -18,12 +18,12 @@ spec:
       description: The Snapshot in JSON format
     - name: iibServiceConfigSecret
       type: string
-      description: Secret containing IIB service Config
       default: iib-services-config
-    - name: iibFromImageOverwriteCredentials
+      description: Secret containing IIB service Config
+    - name: iibOverwriteFromIndexCredential
       type: string
-      description: Secret with overwrite FromImage credentials to be passed to IIB
       default: iib-overwrite-fromimage-credentials
+      description: Secret with overwrite FromImage credentials to be passed to IIB
     - name: iibServiceAccountSecret
       type: string
       description: Secret containing the credentials for IIB service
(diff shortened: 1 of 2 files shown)


internal-services/catalog/iib-pipeline.yaml
```diff
@@ -19,11 +19,9 @@ spec:
     - name: iibServiceConfigSecret
       type: string
       description: Secret containing IIB service Config
-      default: iib-services-config
```
scoheb, line 22: removing the `default` is desired in this PR?

internal-services/catalog/iib-pipeline.yaml
```diff
@@ -19,11 +19,9 @@ spec:
...
-    - name: iibFromImageOverwriteCredentials
+    - name: iibOverwriteFromIndexCredential
       type: string
       description: Secret with overwrite FromImage credentials to be passed to IIB
-      default: iib-overwrite-fromimage-credentials
```
scoheb, line 26: removing the `default` is desired in this PR?

happybhati (approved):

internal-services/catalog/iib-pipeline.yaml
scoheb: nevermind. I see the reason in the other PR

internal-services/catalog/iib-pipeline.yaml
scoheb: nevermind. I see the reason in the other PR

scoheb (dismissed):

internal-services/catalog/iib-pipeline.yaml
theflockers: Not really. I removed to test and forgot to put them back.

internal-services/catalog/iib-pipeline.yaml
theflockers: same.

internal-services/catalog/iib-pipeline.yaml
```diff
@@ -18,12 +18,12 @@ spec:
       description: The Snapshot in JSON format
     - name: iibServiceConfigSecret
       type: string
-      description: Secret containing IIB service Config
```
happybhati, line 21: @theflockers I think this `description` was removed by mistake.

happybhati (changes requested):

internal-services/catalog/iib-pipeline.yaml
theflockers: It moved two lines down @happybhati .

happybhati: ah yes, I see.

happybhati (approved):

PR #39, opened 2023-07-25
feat(HACBS-2272): add pr checks

mmalina:
This will add the same PR checks that we have
in the release-service-bundles repo:
https://github.com/redhat-appstudio/release-service-bundles

- yamllint for all yaml files
- tknparse for all tasks/pipelines that are changed
- gitlint for all commits in PR

### DIFF
--- a/.github/actions/install-tkn/action.yaml
+++ b/.github/actions/install-tkn/action.yaml
@@ -1,0 +1,11 @@
+name: Install tkn
+runs:
+  using: "composite"
+  steps:
+    - run: |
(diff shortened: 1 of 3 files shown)


mmalina:
The PR workflow is still not triggered. I have a ticket open with Github Support and they say that it's because the workflow is not yet present in the main repo. It's a bit weird, but apparently it should work if I create the PR from a topic branch directly in the main repo and not using a fork. So I will try that.

mmalina:
Closing in favor of this new PR which doesn't use my fork and that way the new workflow is triggered properly: https://github.com/hacbs-release/app-interface-deployments/pull/40

PR #40, opened 2023-07-28
feat(HACBS-2272): add pr checks

mmalina:
This will add the same PR checks that we have
in the release-service-bundles repo:
https://github.com/redhat-appstudio/release-service-bundles

- yamllint for all yaml files
- tknparse for all tasks/pipelines that are changed
- gitlint for all commits in PR

The second commit fixes issues found by yamllint.


.github/actions/install-tkn/action.yaml
```diff
@@ -0,0 +1,11 @@
...
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        curl -LO "https://github.com/tektoncd/cli/releases/download/v${TKN_CLI_VERSION}/tektoncd-cli-${TKN_CLI_VERSION}_Linux-64bit.deb"
+        sudo dpkg -i ./tektoncd-cli-${TKN_CLI_VERSION}_Linux-64bit.deb
+      shell: bash
+      env:
+        TKN_CLI_VERSION: 0.30.1
+
```
johnbieren, line 11: extra new line

.gitlint
```diff
@@ -0,0 +1,15 @@
...
+[title-max-length]
+line-length=72
+
+[body-max-line-length]
+line-length=72
+
+[body-first-line-empty]
+
+[contrib-title-conventional-commits]
+
```
johnbieren, line 15: extra new line

johnbieren (approved):
I didn't go line by line to make sure that no typos were made reformatting the yaml, but lgtm

.github/actions/install-tkn/action.yaml
mmalina: Fixed.

.gitlint
mmalina: Fixed.

.github/actions/install-tkn/action.yaml
```diff
@@ -0,0 +1,10 @@
+name: Install tkn
```
davidmogar, line 1: Why not to use an existing one? https://github.com/marketplace/actions/install-tkn

johnbieren (dismissed):

.github/actions/install-tkn/action.yaml
mmalina: The main reason is that the goal here is to do the same as we do in https://github.com/redhat-appstudio/release-service-bundles . 

And why don't we use the existing action there? We used to, but some time ago we moved to installing it directly, since it's only two lines of code and the community action hadn't been updated in a long time: https://github.com/redhat-appstudio/release-service-bundles/pull/109

Since then we moved it to a local action definition since it is used in two places in that repo. I could run it directly in this repo (without the action definition), but I think it's better to stay consistent and have it this way.

davidmogar: Does the action needs to be updated? I checked the code and the url still seems to be valid and you can customize the version. So why to maintain a new action when there's one providing the same functionality?

The other use case is slightly different because you don't define a new action, but I would still use the same existing action for that.

mmalina: > So why to maintain a new action when there's one providing the same functionality?

Because it's fairly trivial. Depending on somebody else's action always brings risks with it. When the action is provided by github, that gives it some trustworthiness. It's acceptable to use community provided actions if they provide value, but in this case I just don't see the benefit.

But that's my opinion. If you think I should use the action, I can change it. Let me know.

mmalina:
Oops, I accidentally deleted the source branch, so the PR got closed. I pushed it again and reopened the PR - hopefully we can still use it.

.github/actions/install-tkn/action.yaml
davidmogar: My opinion is that if there's an action that provides the feature, we should not reinvent the wheel and just use it. Now, if there's a valid reason for not to do it, that's a different story. I just don't see any benefit on writing our own.

If you have a strong opinion on this, go for it. If not, I'd say is better to use existing stuff. Your call, sir ;)

mmalina: @davidmogar I changed it to use the community action. Please check now.

davidmogar (approved):

PR #47, opened 2023-09-08
feat(HACBS-2228): use application in title of MR

scoheb:
Relevant PR for Release Pipeline: https://github.com/redhat-appstudio/release-service-bundles/pull/194


scoheb:
> @scoheb is there an accompanying PR that adds the param on the release pipeline side? Can you reference it? Otherwise it looks good.

See https://github.com/redhat-appstudio/release-service-bundles/pull/194

PR #54, opened 2023-09-28
chore(RHTAPREL-216): adds extra checks and validations to fileUpdates

theflockers:
This PR adds some checks to make sure the fileUpdates data matches the requirements. The following validations were added:
- check if the `path` parameter contains a valid yaml file
- check if the `replacement` parameter value is a sed expression in `|foo|bar|` format
- check if the affected value/block value had only a single line replaced
- check if the result block has the same number of lines as the one found


internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -118,11 +125,44 @@ spec:
...
+                  replacedCount=$(sed -ne "${startBlock},+${valueSize}p" "${targetFile}" |grep "${replaceStr}"  |wc -l)
+                  if [[ $replacedCount != 1 ]]; then
+                      updateError="Too many lines replaced. Check the replace expression"
+                      break
+                  fi
               done
           done
 
+          if [ -n "${updateError}" ]; then
+          yq -o json <<< "error: ${updateError}" |tee $(results.fileUpdatesInfo.path)
```
scoheb, line 159:
```suggestion
            yq -o json <<< "error: ${updateError}" |tee $(results.fileUpdatesInfo.path)
```

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -118,11 +125,44 @@ spec:
...
+                  replacedBlockLines=$(sed -ne "${startBlock},+${valueSize}p" "${targetFile}" |wc -l)
+                  if [[ $replacedBlockLines != $(( $valueSize +1 )) ]]; then
+                      updateError="Text block size differs from the original"
+                      break
+                  fi
+
+                  # check if only a single line was replaced
+                  replacedCount=$(sed -ne "${startBlock},+${valueSize}p" "${targetFile}" |grep "${replaceStr}"  |wc -l)
+                  if [[ $replacedCount != 1 ]]; then
+                      updateError="Too many lines replaced. Check the replace expression"
```
scoheb, line 152: do you think we can show the user what we attempted to replace? Do we know at this point?

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -118,11 +125,44 @@ spec:
...
+
+                  # get the replace part of "|search|replace|"
+                  replaceStr=$(awk -F"|" '{print $3}' <<< "${replacement}")
+
+                  # when the value is a text block we must make sure
+                  # only a single line was replaced and that the result
+                  # block has the same number of lines as before
+                  replacedBlockLines=$(sed -ne "${startBlock},+${valueSize}p" "${targetFile}" |wc -l)
+                  if [[ $replacedBlockLines != $(( $valueSize +1 )) ]]; then
+                      updateError="Text block size differs from the original"
```
scoheb, line 145: are there any details on what the differences might be? Can we show the user?

scoheb (changes requested):

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: We have both the replace expression and the result. The only issue is that this value would need to be saved in the result. Depending on the size of the output, it would fail the PipelineRun.

internal-services/catalog/file-updates-process-file-updates-task.yaml
theflockers: Maybe the error string could be better. The idea here is: In  case the replace expression was miswritten as for instance `|.*|image: some_image|`, if the value is a block like this:
```
   image: image_a
   image: image_b
   image: image_c
   image: image_d
```
The result would be:
```
  image: some_image
```
The idea is prevent cases like this. If the user wants multiple replacements in the same block the user needs to define different replacements for each line.

Maybe I should add something like: `Too many lines replaced. Check if the expression is not too greedy" and print the expression.

scoheb: yeah that works!

internal-services/catalog/file-updates-process-file-updates-task.yaml
```diff
@@ -36,7 +36,7 @@ spec:
       description: fileUpdates state
   steps:
     - name: perform-updates
-      image: quay.io/hacbs-release/release-utils:06643f9e7c2b6ddf5e095bc735ce7c308793815d
+      image: quay.io/lmendes/release-utils:latest
```
theflockers, line 39: this will be changed one the `release-utils` PR is merged.

PR #55, opened 2023-10-05
feat(RHTAPREL-392): add pipeline to create advisories

johnbieren:


johnbieren:
@davidmogar @mmalina @theflockers this is a WIP draft. I am adding you now so you can get a head start on telling me things you don't like. I am planning to investigate:

- publishing advisory path or name or something as a task result
- moving `get_advisory_type` to release-utils image maybe?
- Not passing the advisory_name and having some gitlab automation do it for us so we can run in parallel

But, please, if you don't like other parts of the code, mention it

internal-services/catalog/create-advisory-pipeline.yaml
```diff
@@ -0,0 +1,41 @@
...
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: advisory
+spec:
+  description: >-
+      Pipeline to push an advisory yaml to a Git repository
+  params:
+    - name: repo
+      type: string
+      description: Git repository to push the advisory to
+      default: "https://gitlab.cee.redhat.com/jbieren/rhtap-advisories.git"
```
johnbieren, line 18: This would change to whatever our production repo is

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,115 @@
...
+          E.g. '{"spec":{"product_id":123,"type":"RHSA"}}'
+    - name: application
+      type: string
+      description: Application being released
+  results:
+    - name: result
+      description: Success if the task succeeds, the error otherwise
+  steps:
+    - name: create-advisory
+      image: quay.io/jbieren/release-utils:ansible_5
```
johnbieren, line 32: This would of course change

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,115 @@
...
+              fi
+              echo "RHEA"
+          }
+
+          function get_advisory_num {
+              # This function will skip many advisory numbers as it is so tied to date. Need to
+              # discover if skipping advisory numbers like this is okay.
+              # There is also a race condition picking the advisory number like this.
+              EXISTING=$(for advisory in $(find data/advisories/*/$(date '+%Y') -name "advisory.yaml") ; do
+                  yq '.metadata.name' $advisory | grep ":$ADVISORY_DAY" || true
```
theflockers, line 86: I would rather pass the $ADVISORY_DAY value  as a parameter of `get_advisory_num`

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,115 @@
...
+              echo "${ADVISORY_DAY}1" # First advisory of the day
+          }
+
+          # This also cds into the git repo
+          git_clone_and_checkout --repository "$(params.repo)" --revision "$REPO_BRANCH"
+
+          PRODUCT_ID=$(jq '.spec.product_id' <<< '$(params.advisory_json)')
+          ADVISORY_DAY=$(date '+%m%d')
+          ADVISORY_DIR="data/advisories/$PRODUCT_ID/$(date '+%Y')" # group advisories by product/year
+          ADVISORY_NUM=$(get_advisory_num)
```
theflockers, line 101: same about passing the $ADVISORY_DAY as a parameter.

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,115 @@
...
+                  return
+              fi
+              echo "RHEA"
+          }
+
+          function get_advisory_num {
+              # This function will skip many advisory numbers as it is so tied to date. Need to
+              # discover if skipping advisory numbers like this is okay.
+              # There is also a race condition picking the advisory number like this.
+              EXISTING=$(for advisory in $(find data/advisories/*/$(date '+%Y') -name "advisory.yaml") ; do
```
mmalina, line 85: There's another instance of `$(date '+%Y')` elsewhere. I think we should only determine the year once and then use that. There's the off chance that the script will run around new year, which is very unlikely, but it would still be cleaner to define the year once.

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,115 @@
...
+                  yq '.metadata.name' $advisory | grep ":$ADVISORY_DAY" || true
+              done | sort -r | head -n 1)
+              if [[ ! -z "$EXISTING" ]] ; then
+                  echo "$((${EXISTING#*:}+1))"
+                  return
+              fi
+              echo "${ADVISORY_DAY}1" # First advisory of the day
```
mmalina, line 92: Why not start with 0? And what happens when there are more than 9 advisories in a day? You would have 10069 and then 10070. And if there was yet one more, it would try to create 10070 again. I know that this mechanism is not finalized yet. Just commenting on the current version.

internal-services/catalog/create-advisory-task.yaml
mmalina: Also, why look at the yaml contents for the name, when you can use the directory name directly? That will always match, right?

internal-services/catalog/create-advisory-task.yaml
johnbieren: Yeah, both things you said were correct in this version (I already had fixed it locally but not pushed it yet). I don't know if this advisory name stuff is going to be in the task or a gitlab ci pipeline, so I haven't been updating it. I still appreciate the comments though and will make sure they are addressed one way or another

johnbieren:
@theflockers @mmalina https://issues.redhat.com/browse/RHTAPREL-643 was created as a followup to the advisory stuff. It will remove all the advisory name stuff, that code will not be staying. https://issues.redhat.com/browse/RHTAPREL-392 was just about creating a minimal advisory flow and this lets us create one (not in parallel) so I am opening this back up for review

internal-services/catalog/create-advisory-task.yaml
johnbieren: fixed

internal-services/catalog/create-advisory-task.yaml
johnbieren: fixed

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,104 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: task-create-advisory
```
mmalina, line 5: Question maybe for @theflockers : Why is there a different naming for the file vs. metadata? i.e. blabla-task vs. task-blabla. Just curious how that happened of if there is a reason for that. (Not saying we need to change it or anything.)

internal-services/catalog/create-advisory-task.yaml
mmalina: I still see two different calls of `date '+%Y'`, but this code will go away, so I guess it's not a big deal.

mmalina:
> @theflockers @mmalina https://issues.redhat.com/browse/RHTAPREL-643 was created as a followup to the advisory stuff. It will remove all the advisory name stuff, that code will not be staying. https://issues.redhat.com/browse/RHTAPREL-392 was just about creating a minimal advisory flow and this lets us create one (not in parallel) so I am opening this back up for review

Personally, I would remove that code already if it's not gonna be used, and use something really simple like epoch for now. I wonder what others think.

johnbieren:
> > @theflockers @mmalina https://issues.redhat.com/browse/RHTAPREL-643 was created as a followup to the advisory stuff. It will remove all the advisory name stuff, that code will not be staying. https://issues.redhat.com/browse/RHTAPREL-392 was just about creating a minimal advisory flow and this lets us create one (not in parallel) so I am opening this back up for review
> 
> Personally, I would remove that code already if it's not gonna be used, and use something really simple like epoch for now. I wonder what others think.

Yeah, but the demo "looks nice" with the incrementing numbers right now so I'd like to keep it as is

internal-services/catalog/create-advisory-task.yaml
johnbieren: Yeah I just followed the others

johnbieren:
@mmalina fixed the two calls to `date '+%Y'` I missed that sorry

internal-services/catalog/create-advisory-task.yaml
mmalina: Yeah, if this thread is still unanswered when the PR is ready, just close it.

mmalina:
> > Yeah, but the demo "looks nice" with the incrementing numbers right now so I'd like to keep it as is

OK. Maybe something for next time or can be discussed in a retrospective - that maybe for POC it makes sense to have the code only in a personal fork or something. But I understand it might be hard to do in some cases - like here if you want to have the tasks ready in the internal cluster.

mmalina (dismissed):

johnbieren:
I removed the repo parameter as I figure the most natural place to provide that is in the data field as `advisory.repo` and we pass the entire `data.advisory` key already, so might as well get it from there

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,102 @@
...
+              done | sort -r | head -n 1)
+              if [[ ! -z "$EXISTING" ]] ; then
+                  echo "${2}$((${EXISTING:9}+1))" # Strip first 9 chars for YYYY:MMDD then add 1
+                  return
+              fi
+              echo "${2}0" # First advisory of the day
+          }
+
+          # This also cds into the git repo
+          git_clone_and_checkout --repository "$(jq -rc '.advisory.repo' $(params.advisory_json))" \
```
mmalina, line 83: According to the description of the `advisory_json` pipeline parameter, this won't have the `advisory` key at the top level. So I believe this should be `jq -rc '.repo'`.

johnbieren: Ah, yes, good catch

johnbieren: I will test it one more time once I have https://github.com/redhat-appstudio/release-service-catalog/pull/245 passing CI before I merge here, but still open for review

internal-services/catalog/create-advisory-task.yaml
scoheb: I do not remember any reason unless I was following some other examples at that time....

theflockers (commented):
Just a consideration before approval. Like I did for `file-updates` you might  want to not fail the Pipeline in any situation - even if it actually failed. 
Instead, you might want to save the error in a result and give it back to the calling Pipeline to decide what to do with it.  That's the only way the `user` can know the reason it failed as the Pipeline only adds things to results on success.

internal-services/catalog/create-advisory-task.yaml
johnbieren: This will be reworked in the followup ticket

internal-services/catalog/create-advisory-task.yaml
johnbieren: I changed the task name to `create-advisory-task` instead of `task-create-advisory`

internal-services/catalog/create-advisory-task.yaml
johnbieren: Tested with latest code here and on https://github.com/redhat-appstudio/release-service-catalog/pull/245 and it works as expected (including the failure scenario)

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,108 @@
...
+          #!/usr/bin/env sh
+          set -ox pipefail
+
+          # This should get overwritten if any command fails
+          echo "Success" > $(results.result.path)
+
+          failfunc() {
+              echo "Failure" > $(results.result.path)
+          }
+          trap "failfunc" ERR
```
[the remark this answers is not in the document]
johnbieren: Updated per offline conversation

theflockers (dismissed):

internal-services/catalog/create-advisory-pipeline.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - name: advisory_json
+      type: string
+      description: |
+          String containing a JSON representation of the advisory data. It should not contain the `advisory`
+          top level key
```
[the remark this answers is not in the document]
johnbieren: Okay, I moved it up from the line below

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,115 @@
...
+      Pushes an advisory yaml to a Git repository
+  params:
+    - name: repo
+      type: string
+      description: Git repository to push the advisory to
+    - name: advisory_json
+      type: string
+      description: |
+          String containing a JSON representation of the advisory data. It should not contain the `advisory`
+          top level key
```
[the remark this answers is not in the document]
johnbieren: Okay, I moved it up from the line below

internal-services/catalog/create-advisory-task.yaml
mmalina: Hmm. The pipeline has pipeline in filename, but not name, but the task has task in both. Do we need it in the task name in the yaml? Probably not a big deal, since this means it will not be consistent with the rest of the repo anyway.

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+          # This should get overwritten if any command fails
+          echo -n "Success" > $(results.result.path)
+
+          failfunc() {
+              local err=$1
+              local line=$2
+              local command="$3"
+              echo "$0: ERROR '$command' failed at line $line - exited with status $err" > $(results.result.path)
+              exit 0 # exit the script cleanly as there is no point in proceeding past an error
```
mmalina, line 67: > as there is no point in proceeding past an error

Also below:

> the task should never fail with nonzero exit code

But why? I know the answer - because the internal request is called from the main RHTAP cluster and we want to be able to see the result there. Can you maybe explain that briefly somewhere? Either in one of the comments, or in the task description - that it will return 0 even if something fails and why.

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,108 @@
...
+          valueFrom:
+            secretKeyRef:
+              name: create-advisory-secret
+              key: git_author_email
+      script: |
+          #!/usr/bin/env sh
+          set -ox pipefail
+
+          # This should get overwritten if any command fails
+          echo "Success" > $(results.result.path)
```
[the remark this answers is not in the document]
mmalina: For the comment above - I wrote it 5 days ago, but I think it only got published now (was Pending). Since you changed the result text to provide more details, that part is no longer relevant. But whatever we save, it should probably still not contain a newline I'd say.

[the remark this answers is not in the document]
johnbieren: Added `-n` to the `echo`s to the result

johnbieren:
@mmalina It won't me reply directly to the comment about the task name and filename, but yeah I just copied what we had in the repo, so if you are okay with it, I'll just leave it for now

internal-services/catalog/create-advisory-task.yaml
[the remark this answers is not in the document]
mmalina: Yep, awesome!

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -0,0 +1,115 @@
...
+            secretKeyRef:
+              name: create-advisory-secret
+              key: git_author_email
+      script: |
+          #!/usr/bin/env sh
+          set -eox pipefail
+
+          REPO_BRANCH=main
+
+          # loading git and gitlab functions
```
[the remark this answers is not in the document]
mmalina: Yeah, when they were added, I opposed to the idea that one of the functions file would call an init function upon loading (with `source` or `.`). So that was removed and we now call the init function separately. But sourcing the function files in the Dockerfile always would probably be fine.

internal-services/catalog/create-advisory-task.yaml
mmalina: Looks good, thanks!

internal-services/catalog/create-advisory-task.yaml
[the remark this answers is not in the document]
mmalina: (But I think this is definitely out of scope for this PR. @davidmogar if you feel this should be changed, feel free to create a story for it.)

mmalina (approved):

happybhati (approved):

PR #60, opened 2023-10-31
refactor(RHTAPREL-643): update create-advisory task logic

johnbieren:
This commit changes the create-advisory task to push the advisories to a staging directory. The advisory name is also just a uuidgen now, as a gitlab ci pipeline will do that part instead. The logic that was present to help determine the advisory name is removed.

In addition, the error and trap handling in the task was updated

### DIFF
--- a/internal-services/catalog/create-advisory-task.yaml
+++ b/internal-services/catalog/create-advisory-task.yaml
@@ -58,19 +58,21 @@ spec:
               key: git_author_email
       script: |
           #!/usr/bin/env sh
-          set -ox pipefail
+          set -eo pipefail
 
-          # This should get overwritten if any command fails
-          echo -n "Success" > $(results.result.path)
-
-          failfunc() {
+          exitfunc() {
               local err=$1
               local line=$2
               local command="$3"
-              echo -n "$0: ERROR '$command' failed at line $line - exited with status $err" > $(results.result.path)
-              exit 0 # exit the script cleanly as there is no point in proceeding past an error
+              if [ $err -eq 0 ] ; then
+                  echo -n "Success" > $(results.result.path)
+              else
+                  echo -n "$0: ERROR '$command' failed at line $line - exited with status $err" > $(results.result.path)
+              fi
+              exit 0 # exit the script cleanly as there is no point in proceeding past an error or exit call
           }
-          trap 'failfunc $? $LINENO "$BASH_COMMAND"' ERR # the task should never fail with nonzero exit code
+          # due to set -e, this catches all EXIT and ERR calls and the task should never fail with nonzero exit code
+          trap 'exitfunc $? $LINENO "$BASH_COMMAND"' EXIT
 
           REPO_BRANCH=main
           REPO=$(jq -r '.repo' <<< '$(params.advisory_json)')
--- a/internal-services/catalog/create-advisory-task.yaml
+++ b/internal-services/catalog/create-advisory-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: create-advisory-task
   labels:
-    app.kubernetes.io/version: "0.1"
+    app.kubernetes.io/version: "0.2"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
@@ -34,7 +34,7 @@ spec:
       description: Success if the task succeeds, the error otherwise
   steps:
     - name: create-advisory
-      image: quay.io/hacbs-release/release-utils:518c0423a3b3502c7b9ef6fd8652995bec2b4e1a
+      image: quay.io/hacbs-release/release-utils:6fe50b65600418f9b046c32733fc896a4df02d41
       env:
         - name: GITLAB_HOST
           valueFrom:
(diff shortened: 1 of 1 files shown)


internal-services/catalog/create-advisory-task.yaml
```diff
@@ -34,7 +34,7 @@ spec:
       description: Success if the task succeeds, the error otherwise
   steps:
     - name: create-advisory
-      image: quay.io/hacbs-release/release-utils:518c0423a3b3502c7b9ef6fd8652995bec2b4e1a
+      image: quay.io/hacbs-release/release-utils:518c0423a3b3502c7b9ef6fd8652995bec2b4e1a # needs update
```
mmalina, line 37: Why does this need an update? Just to have a new image, or is there something specific that is missing? Maybe the `git_push_with_retries` function is new? Can add a link to the related PR to the description?

internal-services/catalog/create-advisory-task.yaml
mmalina, line 1: WDYT about changing the version to 0.2? It seems most tasks/pipelines in this repo are on version 0.1, only this one is on 0.2: https://github.com/hacbs-release/app-interface-deployments/blob/main/internal-services/catalog/iib-pipeline.yaml

So it seems we mostly don't care about versions here, but I think why not upversion?

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -81,35 +81,18 @@ spec:
...
-          ADVISORY_FILEPATH="${ADVISORY_DIR}/${ADVISORY_NUM}/advisory.yaml"
-          mkdir -p "${ADVISORY_DIR}/${ADVISORY_NUM}"
+          ADVISORY_DIR="staging/advisories/$(params.origin)/$(date '+%Y')" # group advisories by <origin workspace>/year
+          mkdir -p "${ADVISORY_DIR}"
+          ADVISORY_FILEPATH="${ADVISORY_DIR}/$(uuidgen).yaml"
 
           # Create advisory file
           /home/utils/apply_template.py -o $ADVISORY_FILEPATH --template /home/templates/advisory.yaml.jinja \
-          --data '{"advisory_name":"'$ADVISORY_NAME'","advisory":$(params.advisory_json)}'
+          --data '{"advisory":$(params.advisory_json)}'
```
mmalina, line 93: I assume `advisory_name` will be injected by the Gitlab CI pipeline that will generate the name, right?

internal-services/catalog/create-advisory-task.yaml
johnbieren: Yes, it will need the `git_push_with_retries` function

johnbieren:
Depends on https://github.com/redhat-appstudio/release-service-utils/pull/78

internal-services/catalog/create-advisory-task.yaml
johnbieren: yep. The CI pipeline will move it from staging to data and set the advisory name (both the filename and `metadata.name`)

internal-services/catalog/create-advisory-task.yaml
johnbieren: good call, bumped version

internal-services/catalog/create-advisory-task.yaml
davidmogar: Yep, it will. The name needs to have a numeration that increases by 1 with each new advisory, so it needs to be done in the CI itself to avoid race conditions.

internal-services/catalog/create-advisory-task.yaml
mmalina: cool

internal-services/catalog/create-advisory-task.yaml
mmalina: ok, cool - I got to that second PR just after reviewing this one :)

mmalina (approved):

mmalina (dismissed):

johnbieren:
I updated the task error handling so this PR is two commits now. I didn't like that the error returned to the IR was about the checkout when, for example, the git clone fails as `git_clone_and_checkout` would execute completely even if part of it failed

internal-services/catalog/create-advisory-task.yaml
mmalina, line 1: I'm fine with the changes. But wouldn't it be enough to just add the `set -e` (as you did)? I would say that the rest of the changes was not needed, no?

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -58,19 +58,20 @@ spec:
...
               local err=$1
               local line=$2
               local command="$3"
-              echo -n "$0: ERROR '$command' failed at line $line - exited with status $err" > $(results.result.path)
+              if [ $err -eq 0 ] ; then
+                  echo -n "Success" > $(results.result.path)
+              else
+                  echo -n "$0: ERROR '$command' failed at line $line - exited with status $err" > $(results.result.path)
+              fi
               exit 0 # exit the script cleanly as there is no point in proceeding past an error
```
mmalina, line 75: If we go with catching EXIT, this comment should be updated - it is now called on either successful finish or on a failure.

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -58,19 +61,20 @@ spec:
               key: git_author_email
       script: |
           #!/usr/bin/env sh
-          set -ox pipefail
+          set -eox pipefail
```
mmalina, line 64: The `x` should probably be removed as it will leak the `ACCESS_TOKEN` when using the env var in the git functions.

internal-services/catalog/create-advisory-task.yaml
johnbieren: No because we don't want the task to fail. When I had `set -e`, it did

internal-services/catalog/create-advisory-task.yaml
johnbieren: will update

internal-services/catalog/create-advisory-task.yaml
johnbieren: When I did that, `git_push_with_retries` failed the task due to the check of if `attempt -gt RETRIES`

internal-services/catalog/create-advisory-task.yaml
johnbieren: fixed

internal-services/catalog/create-advisory-task.yaml
johnbieren: fixed, thanks

johnbieren:
I'd like to pull https://github.com/redhat-appstudio/release-service-utils/pull/81 into this too

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -58,19 +61,21 @@ spec:
...
+              if [ $err -eq 0 ] ; then
+                  echo -n "Success" > $(results.result.path)
+              else
+                  echo -n "$0: ERROR '$command' failed at line $line - exited with status $err" > $(results.result.path)
+              fi
+              exit 0 # exit the script cleanly as there is no point in proceeding past an error or exit call
           }
-          trap 'failfunc $? $LINENO "$BASH_COMMAND"' ERR # the task should never fail with nonzero exit code
+          trap 'exitfunc $? $LINENO "$BASH_COMMAND"' EXIT # due to set -e, this catches all EXIT and ERR calls and the
+                                                          # task should never fail with nonzero exit code
```
mmalina, line 78: Nit: If the comment is too long to fit on the same line, I would suggest putting it above the command (and use as many lines as needed) - adding an empty line below with a comment at the end looks strange.

johnbieren: great idea. I hate how it looks now :slightly_smiling_face:

johnbieren: fixed

mmalina: Awesome, thanks :)

mmalina (dismissed):

johnbieren:
Last update to this, changed the image to the `redhat-appstudio` one and one that has the change to utils to not remove undefined variables

happybhati (approved):

PR #69, opened 2023-11-27
fix(RHTAPBUGS-1019): add support for staged index use case

scoheb:
- set parameters accordingly when stagedIndex = true.
- we want to force opt_in to false and ensure no images are overwritten or published

Related to https://github.com/redhat-appstudio/release-service-catalog/pull/305

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.1"
+    app.kubernetes.io/version: "0.2.0"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
@@ -49,6 +49,10 @@ spec:
       type: string
       default: "false"
       description: Whether this build is a hotfix build
+    - name: stagedIndex
+      type: string
+      default: "false"
+      description: Whether this build is for a staged index build
   results:
     - name: jsonBuildInfo
       description: JSON build information for the requested build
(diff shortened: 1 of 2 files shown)


happybhati (approved):

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -131,13 +135,19 @@ spec:
...
             mustPublishIndexImage="true"
         else
-            echo "Fragment has \`fbc_opt_in==${fbcOptIn}\`"
-            # in case of more than one image all should have the same fbc_opt_in value
-            if [ $(wc -w <<< ${fbcOptIn[@]}) -ne 1 ]; then
-                fbcOptIn="false"
-                mustSignIndexImage="false"
-                mustPublishIndexImage="false"
-            fi
+            if [ "$(params.stagedIndex)" == "true" ]; then
```
mmalina, line 138: Please give it a thought as well, but I think we can remove two levels of nesting like this:
```sh
if [ "$(params.hotfix)" == "true" ]
...
elif [ "$(params.stagedIndex)" == "true" ]
...
elif [ $(wc -w <<< ${fbcOptIn[@]}) -ne 1 ]
...
fi
```
And then perhaps add the echo after all of that:
```sh
echo "Fragment has \`fbc_opt_in==${fbcOptIn}\`"
```

mmalina:
> Requires [redhat-appstudio/release-service-catalog#305](https://github.com/redhat-appstudio/release-service-catalog/pull/305)

Since the new param is optional and has a default, this doesn't really require the other one. It's rather the opposite - this one is needed by 305, right? It's a detail, but I was both of them saying they require the other which seemed odd.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.1"
+    app.kubernetes.io/version: "0.1.1"
```
mmalina, line 7: It's a new functionality, so probably 0.2.0?

scoheb: fixed

theflockers (approved):

PR #73, opened 2024-02-15
feat(RHTAPREL-793): create-advisory task injects signing key

johnbieren:
This commit modifies the create-advisory task and pipeline to receive a configmap name and fetch the signing key from it to inject into the content.images section of the advisory yaml.


johnbieren:
Draft since I haven't tested this yet. However, I opened it and tagged you Martin in case you have feedback on the `jq` calls

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -93,9 +95,14 @@ spec:
           mkdir -p "${ADVISORY_DIR}"
           ADVISORY_FILEPATH="${ADVISORY_DIR}/$(uuidgen).yaml"
 
+          # Inject signing key into advisory_json
+          signingKey=$(kubectl get configmap $(params.config_map_name) -o jsonpath="{.data.SIG_KEY_NAME}"
+          advisoryJsonWithKey=$(jq -c --arg key "$signingKey" \
+            '.content.images |= map(. + {"signingKey": $signingKey})' <<< $(params.advisory_json)
```
mmalina, line 101: The intention is to end up with ".content.images.signingKey" included in the json, right? Also, you called the arg `key`, so you need to use it in the query.

```suggestion
            '.content.images += {"signingKey": $key}' <<< $(params.advisory_json)
```

johnbieren: To have the signingKey in every entry of `.content.images[]`
Nice catch

johnbieren:
```
jq --arg key "$key" '.images += {"signingKey": $key}' <<< $json
jq: error (at <stdin>:64): array ([{"containe...) and object ({"signingKe...) cannot be added
```

```
jq --arg key "$key" '.images |= map(. + {"signingKey": $key})' <<< $json
{
  "images": [
    {
      "architecture": "amd64",
      "signingKey": "foobar",
      }
    },
    {
      "architecture": "ppc64le",
      "signingKey": "foobar",
    }
  ]
}
```

mmalina: I see. Then how about this?
```
jq --arg key "mykey" '.content.images[] += {"signingKey": $key}' <<< '{"content": {"bla": "bla", "images": [{"name": "img1"}, {"name": "img2"}]}}'
{
  "content": {
    "bla": "bla",
    "images": [
      {
        "name": "img1",
        "signingKey": "mykey"
      },
      {
        "name": "img2",
        "signingKey": "mykey"
      }
    ]
  }
}
```

johnbieren: That seems to work. Updated!

mmalina: Cool!

happybhati (approved):

mmalina (approved):

PR #74, opened 2024-02-20
fix(KFLUXBUGS-1116): add the ability to get the manifest digests

scoheb:
- we need to be able to access the registry-proxy location to be able to skopeo inspect to obtain the digest.
- this will relay that digest list as a result.


internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -243,6 +245,15 @@ spec:
             echo "Timeout while waiting for the build to finish"
             echo "Build timeout" < $(results.buildState.path)
         fi
+
+        # get the manifest digests
+        indexImageCopy=`cat $(results.buildState.path) | jq -cr  .internal_index_image_copy`
```
johnbieren, line 250: looks like two spaces after `cr`

johnbieren (approved):

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -243,6 +245,15 @@ spec:
...
+        # Use this to obtain the manifest digests for each arch in manifest list
+        indexImageDigestsRaw=$(skopeo inspect --raw docker://$indexImageCopy)
+        indexImageDigests=$(echo ${indexImageDigestsRaw} | \
+           jq -r '.manifests[] | select(.mediaType=="application/vnd.docker.distribution.manifest.v2+json") | .digest')
```
mmalina, line 254: What's the intention of the select? Do you expect some other stuff in the manifest list?

More importantly, this seems to break for images without a manifest list. Or can we safely assume that from now on all index images will always be multi-arch with a manifest list?

Here's an example of an image with `"mediaType": "application/vnd.oci.image.manifest.v1+json"`:
```
skopeo inspect --raw docker://quay.io/hacbs-release-tests/dev-release-team-tenant/simple-python/devfile-sample-python-basic@sha256:c999142ffd747c3783bd3b20c9092da6cc3931625cf13f200536e62c3cc67a8e| jq -r '.manifests[] | select(.mediaType=="application/vnd.docker.distribution.manifest.v2+json") | .digest'
jq: error (at <stdin>:0): Cannot iterate over null (null)
```

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
scoheb: fixed

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
scoheb: > What's the intention of the select? Do you expect some other stuff in the manifest list?
I am not sure to be honest. Seems safe as those digests are the ones I want to sign.

> More importantly, this seems to break for images without a manifest list. 
Added protection for that case

> Or can we safely assume that from now on all index images will always be multi-arch with a manifest list?
Confirmed with IIB team. we can safely assume `all index images will always be multi-arch with a manifest list`
Also added a note.

mmalina: > > What's the intention of the select? Do you expect some other stuff in the manifest list?
> 
> I am not sure to be honest. Seems safe as those digests are the ones I want to sign.

Cool, I was just curious, sounds good.

> > More importantly, this seems to break for images without a manifest list.
> 
> Added protection for that case

Great, thanks!

> > Or can we safely assume that from now on all index images will always be multi-arch with a manifest list?
> 
> Confirmed with IIB team. we can safely assume `all index images will always be multi-arch with a manifest list` Also added a note.

Good to hear that.

mmalina (approved):

PR #80, opened 2024-03-04
chore(noticket): promote main to stable

theflockers:
Promotes main branch to stable. Contains commits for the following ticket:

- RHTAPREL-858
- RHTAPREL-815
- RHTAPREL-793


johnbieren (approved):
Kind of not great that this is how we do this IMO

theflockers:
> Kind of not great that this is how we do this IMO

Agree @johnbieren. Anyhow we need a bit of care with this repo. I think we have a ticket to improve it, we need to prioritize it.

PR #81, opened 2024-03-06
chore(RHTAPREL-859): bump image used in create-advisory-task

johnbieren:


mmalina (commented):
Can you give at least some context? Either a link to a PR for the utils image, or at least mention that you needed this to get some specific update in the image. Looking at the jira, I assume it's the template update?

johnbieren:
> Can you give at least some context? Either a link to a PR for the utils image, or at least mention that you needed this to get some specific update in the image. Looking at the jira, I assume it's the template update?

Yeah, that's it. I will update the commit, good suggestion. I didn't update the task version string since it is just an image bump, but yours which adds the shipDate thing should IMO since that will also add a little to the task script

mmalina (approved):

PR #82, opened 2024-03-06
feat(RHTAPREL-799): add ship date to advisory

mmalina:
The advisory template now expects a new variable
to set metadata.ship_date in the advisory.

Depends on: https://github.com/redhat-appstudio/release-service-utils/pull/129

### DIFF
--- a/internal-services/catalog/create-advisory-task.yaml
+++ b/internal-services/catalog/create-advisory-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: create-advisory-task
   labels:
-    app.kubernetes.io/version: "0.7"
+    app.kubernetes.io/version: "0.8"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
@@ -41,7 +41,7 @@ spec:
       description: The advisory url if the task succeeds, empty string otherwise
   steps:
     - name: create-advisory
-      image: quay.io/redhat-appstudio/release-service-utils:d315777adea2aff34cd0ab28a85a139cae2f84b7
+      image: quay.io/redhat-appstudio/release-service-utils:8106fea3af80b7978c599dd9fc8cee0a679f3f35
       env:
         - name: GITLAB_HOST
           valueFrom:
@@ -100,7 +100,9 @@ spec:
           # This will be replaced in RHTAPREL-811
           ADVISORY_NUM=$(((RND=RANDOM<<15|RANDOM)) ; echo ${RND: -6})
 
-          YEAR=$(date '+%Y')
+          # Use ISO 8601 format in UTC/Zulu time, e.g. 2024-03-06T17:27:38Z
+          SHIP_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+          YEAR=${SHIP_DATE%%-*} # derive the year from the ship date
           # group advisories by <origin workspace>/year
           ADVISORY_DIR="data/advisories/$(params.origin)/${YEAR}/${ADVISORY_NUM}"
           mkdir -p "${ADVISORY_DIR}"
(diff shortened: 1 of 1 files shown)


internal-services/catalog/create-advisory-task.yaml
```diff
@@ -41,6 +41,7 @@ spec:
       description: The advisory url if the task succeeds, empty string otherwise
   steps:
     - name: create-advisory
+      # TODO: Update this once https://github.com/redhat-appstudio/release-service-utils/pull/129 is merged
```
mmalina, line 44: Adding a note here to update the image before merging this.

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -101,6 +102,8 @@ spec:
           ADVISORY_NUM=$(((RND=RANDOM<<15|RANDOM)) ; echo ${RND: -6})
 
           YEAR=$(date '+%Y')
+          # Use ISO 8601 format in UTC/Zulu time, e.g. 2024-03-06T17:27:38Z
+          SHIP_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
```
johnbieren, line 106: Should $YEAR then be extracted from this? To prevent the super super rare edge case? :sunglasses:

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -112,9 +115,14 @@ spec:
           advisoryJsonWithKey=$(jq -c --arg key "$signingKey" \
             '.content.images[] += {"signingKey": $key}' <<< '$(params.advisory_json)')
 
+          # Prepare variables for the advisory template
+          DATA=$(jq -c '{"advisory":{"spec":.}' <<< $advisoryJsonWithKey)
+          DATA=$(jq -c --arg advisory_name "$ADVISORY_NAME" --arg advisory_ship_date "$SHIP_DATE" \
```
johnbieren, line 120: If this actually works, it looks cool

internal-services/catalog/create-advisory-task.yaml
mmalina: I had a similar concern with the year, but there the gap between defining the year and actually pushing the commit could definitely be bigger, because you're calling an api. Here it's really just milliseconds, so I'd say it's ok. 

But just for you... done 😎

internal-services/catalog/create-advisory-task.yaml
mmalina: Yeah, I started adding it to the command below and it was getting very long, so I came up with this. The first one just wraps the `$advisoryJsonWithKey` json with advisory.spec and the second adds the two other properties.

internal-services/catalog/create-advisory-task.yaml
johnbieren: :joy:

internal-services/catalog/create-advisory-task.yaml
johnbieren: Sweet, I've never seen the `$ARGS.named` syntax

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -112,9 +115,14 @@ spec:
...
+            '$ARGS.named + .' <<< $DATA)
+
           # Create advisory file
-          /home/utils/apply_template.py -o $ADVISORY_FILEPATH --template /home/templates/advisory.yaml.jinja \
-            --data '{"advisory_name":"'${ADVISORY_NAME}'","advisory":{"spec":'"${advisoryJsonWithKey}"'}}'
+          /home/utils/apply_template.py -o $ADVISORY_FILEPATH --data $DATA \
+            --template /home/templates/advisory.yaml.jinja \
```
johnbieren, line 125: extra `\` at the end

internal-services/catalog/create-advisory-task.yaml
mmalina: Me neither. I saw it randomly when googling how to compose a json from scratch using jq.

internal-services/catalog/create-advisory-task.yaml
mmalina: Good catch. Fixed.

internal-services/catalog/create-advisory-task.yaml
johnbieren: I will be testing the advisory stuff end to end soon enough anyways so we will see how it goes :slightly_smiling_face:

johnbieren (dismissed):
Approved, obviously needs to hold on the image though

internal-services/catalog/create-advisory-task.yaml
mmalina: https://github.com/redhat-appstudio/release-service-utils/pull/129 is merged, so I updated the image.

johnbieren (approved):

PR #88, opened 2024-03-20
feat(RHTAPREL-878): update create-advisory-task secrets

johnbieren:
The create-advisory-task secrets were moved on the cluster and the
ERRATA_API should be a secret so the stage and prod tasks don't differ.


davidmogar (changes requested):
Please, add a description to both the PR and the commit. Also, this is a feature not a refactor.

theflockers (approved):
LGTM after David's comments are resolved.

davidmogar (approved):

PR #98, opened 2024-04-09
feat(RHTAPREL-663): add embargo check pipeline

johnbieren:
Add both a tekton pipeline and task for checking if any CVEs are embargoed.

### DIFF
--- a/internal-services/catalog/check-embargoed-cves-task.yaml
+++ b/internal-services/catalog/check-embargoed-cves-task.yaml
@@ -1,0 +1,89 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: check-embargoed-cves-task
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: release
+spec:
+  description: |
+      Checks if any passed CVEs are embargoed.
+      The task will always exit 0 even if something fails. This is because the task result will not be
+      set if the task fails, and the task result should always be set and propagated back to the cluster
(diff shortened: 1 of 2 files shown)


internal-services/catalog/check-embargoed-cves-task.yaml
```diff
@@ -0,0 +1,89 @@
...
+              echo "Checking CVE ${CVE}"
+              # Get token. They are short lived, so get one for before each request
+              TOKEN=$(curl --retry 3 --negotiate -u : ${OSIDB_URL}/auth/token | jq -r '.access')
+              EMBARGOED=$(curl --retry 3 -H 'Content-Type: application/json' -H "Authorization: Bearer ${TOKEN}" \
+                  "${OSIDB_URL}/osidb/api/v1/flaws?cve_id=${CVE}&include_fields=cve_id,embargoed" \
+                  | jq .results[0].embargoed)
+              # null would mean no access to the CVE, which may mean embargoed, and true means embargoed
+              if [ $EMBARGOED != "false" ] ; then
+                  echo "CVE ${CVE} is embargoed"
+                  echo -n "${CVE} " >> $(results.embargoed_cves.path)
```
theflockers, line 85: Is there any possibility of the list for CVEs being bigger than the result size? Asking just in case.

johnbieren: hmm. They are like 13 characters each, so in theory eventually? But I'd say if we end up with a customer fixing 1300 CVEs in one release, let's deal with it then

johnbieren: Good thinking though

theflockers (approved):

happybhati (approved):

PR #99, opened 2024-04-10
fix(RHTAPREL-891): declare param as env var in advisory task

johnbieren:
In order to support string values with single quotes in the advisory_json parameter of the create-advisory-task, the parameter is declared as an env variable for the task script instead of using the parameter directly.


johnbieren:
See (already merged) https://github.com/hacbs-release/app-interface-deployments/pull/97

PR #102, opened 2024-04-23
chore: sync stable with main

scoheb:
- updates for advisory task
- add embargo check pipeline
- support prod-p01


happybhati (approved):

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -81,6 +81,8 @@ spec:
             secretKeyRef:
               name: errata-service-account
               key: base64_keytab
+        - name: "ADVISORY_JSON"
```
johnbieren, line 84: Something is off with this diff because this line already exists in stable https://github.com/hacbs-release/app-interface-deployments/blob/stable/internal-services/catalog/create-advisory-task.yaml#L84

PR #111, opened 2024-05-21
fix(RELEASE-928): switch to using secret for repo

scoheb:
- remove the need to use the repo value from ReleaseServiceConfig
- use the repo field from the Advisory Secret


johnbieren (approved):
Not sure how strict we are about updating the version strings in this repo, but otherwise lgtm

PR #125, opened 2024-06-07
fix: renaming the role back to internal-services-manager-rolebinding

theflockers:
This PR renames the ClusterRole and RoleBinding with the prefix `internal-services`.


jinqi7 (approved):

theflockers:
not required.

PR #127, opened 2024-06-12
feat(RELEASE-951): integrate Checkton for linting

seanconroy2021:
This updates the linting workflow by adding Checkton. Checkton will review the code for linting errors.

Changes:
* Added a step to the workflow to run Checkton on PR.
* Configured Checkton to upload the SARIF report.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -50,3 +50,24 @@ jobs:
         run: python -m pip install gitlint
       - name: Run gitlint check
         run: gitlint --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
+  checkton:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+      # Differential Checkton requires full git history
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Run Checkton
+        id: checkton
+        uses: chmeliik/checkton@v0.1.2 # Migrating to the konflux-ci org
+        with:
+         # Let there be green. GitHub's code scanning will do the reporting.
+          fail-on-findings: false
+          find-copies-harder: true
+      - name: Upload SARIF File
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: ${{ steps.checkton.outputs.sarif }}
+          # Avoid clashing with ShellCheck
+          category: checkton


.github/workflows/lint.yaml
```diff
@@ -50,3 +50,20 @@ jobs:
...
+  checkton:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Run Checkton
+        id: checkton
+        uses: chmeliik/checkton@main
```
seanconroy2021, line 62: I want to tag this as a version, but the action/repo has no versioning available. We could run into a breakage if the repo is updated.

davidmogar: True... In this case I wouldn't make the task mandatory maybe... I would also ask the author (though an issue in their repo) to tag the versions.

seanconroy2021: Created an [Issue](https://github.com/chmeliik/checkton/issues/12)

.github/workflows/lint.yaml
```diff
@@ -50,3 +50,20 @@ jobs:
...
+        with:
+          fail-on-findings: false
+      - name: Upload SARIF File
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: ${{ steps.checkton.outputs.sarif }}
+          category: checkton
```
theflockers, line 69: please remove extra line.

seanconroy2021: @theflockers, I don't see any extra line or weird character. I may have fixed it previously.

seanconroy2021: fixed

.github/workflows/lint.yaml
mmalina, line 1: Can you say in the commit/PR description that this is based on https://github.com/konflux-ci/build-definitions/pull/1051 ? Also, I find the comments in the original one useful, could we have them as well?

.github/workflows/lint.yaml
```diff
@@ -50,3 +50,20 @@ jobs:
         run: python -m pip install gitlint
       - name: Run gitlint check
         run: gitlint --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
+  checkton:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
mmalina, line 59: In the original [PR](https://github.com/konflux-ci/build-definitions/pull/1051/files) this has a comment "Differential Checkton requires full git history" - but is that really true? Shouldn't one level be enough for diff?

.github/workflows/lint.yaml
```diff
@@ -50,3 +50,20 @@ jobs:
...
+      - name: Run Checkton
+        id: checkton
+        uses: chmeliik/checkton@v0.1.2
+        with:
```
mmalina, line 63: The original [PR](https://github.com/konflux-ci/build-definitions/pull/1051/files) also sets `find-copies-harder: true`. Did we not want that?

More importantly, the acceptance criteria in Jira say that we should fix all the issues that the tool finds. But by default it uses differential mode, so only new code is checked. So we don't know if there are issues in the existing code.

.github/workflows/lint.yaml
mmalina, line 1: Do we know why the build team actually never merged the check into their repo? It's not here: https://github.com/konflux-ci/build-definitions/tree/main/.github/workflows

.github/workflows/lint.yaml
seanconroy2021: The Checkton tool is designed to check only new code by default. In a  Konflux demo call, the owner demonstrated this feature and explained that this decision was made because linting the entire codebase would result in too many issues. If we want to lint the entire codebase, we would need to use a different tool along with Checkton.

`find-copies-harder: true`. Did we not want that?  We can add that if you would like ?

.github/workflows/lint.yaml
seanconroy2021: From what I know, the owner, who is part of the build team, wants to move Checkton to Konflux-ci, which is why the PR hasn't been merged yet. I was talking in Slack about this, and we decided to move forward and update the reference later. https://redhat-internal.slack.com/archives/C04J47GL936/p1720170153614879

seanconroy2021: Yea will add it

.github/workflows/lint.yaml
seanconroy2021: Added in the  `find-copies-harder: true`

mmalina: > In a Konflux demo call, the owner demonstrated this feature and explained that this decision was made because linting the entire codebase would result in too many issues.

OK, but in our jira we specifically said we did want to fix all the issues it finds in our current code. And it can be accomplished by setting differential mode to false - it could be a separate PR where you temporarily set that to see what needs fixing, fix all the issues and eventually remove the differential=false. I think it would be worth to at least see how many issues it would find and then we can decide if we want to fix all of those or not.

.github/workflows/lint.yaml
mmalina: ok, good to know, thanks.

mmalina: Thanks.

.github/workflows/lint.yaml
seanconroy2021: Got it, I see what you mean now. Here's the link to the PR with the linted code problems: https://github.com/seanconroy2021/app-interface-deployments/pull/2. There are only a few issues, so I think it makes sense to include them here.

mmalina: Cool, thanks. But why does that PR show 23 commits? Either you'll need to update your fork's main, but you could also create that PR against the main repo's main. And keep it as Draft (and perhaps with some "Test" prefix) so that it's not accidentally merged.

And I agree - it actually seems it's just the one thing repeated in several places?
````sh
export KRB5_CONFIG=`mktemp`
```
I'm not exactly sure what the warning is saying though. I would start by changing it to `export KRB5_CONFIG=$(mktemp)` and maybe that will fix it.

So you can add a second commit to this PR with that fix (and reference that testing PR so that anybody can check and see why those things are being fixed).

.github/workflows/lint.yaml
seanconroy2021: I have tested it with 'one level ', and it works. I don't mind removing it; I'm just following the [docs for the action.](https://github.com/chmeliik/checkton) What do you think is best? 👍🏻

mmalina: Oh, now I realize `fetch-depth: 0` may actually be really needed - that's because a PR can contain multiple commits and so with just one level deep you might not be able to do the diff. So yeah, let's keep this.

mmalina (approved):

PR #133, opened 2024-06-12
feat(RELEASE-977): bump task base images

johnbieren:


happybhati (approved):

mmalina (commented):
I know it's in the jira, but maybe just mention here that it's about migrating to the other quay org, like you do in the other PR?

johnbieren:
> I know it's in the jira, but maybe just mention here that it's about migrating to the other quay org, like you do in the other PR?

Updated commit body

PR #137, opened 2024-06-28
fix: increase verbosity

theflockers:
this commit increases the verbosity of the `iib-add-fbc-fragment-to-index-image-task`
by adding a `set -x` to it.

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -116,6 +116,7 @@ spec:
 
         /usr/bin/kinit -V $(cat /mnt/service-account-secret/principal) -k -t /tmp/keytab
 
+        set -x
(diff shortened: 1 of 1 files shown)


internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -116,7 +116,8 @@ spec:
 
         /usr/bin/kinit -V $(cat /mnt/service-account-secret/principal) -k -t /tmp/keytab
 
-        # check if this fbc fragment is opt-in
+        set -x
+        # chek if this fbc fragment is opt-in
```
johnbieren, line 120: unintended change with typo?

theflockers: not even noticed. Thanks for mentioning!

johnbieren (approved):

PR #140, opened 2024-07-09
feat(RELEASE-971): add task and pipeline to push disk images

johnbieren:
This task adds a pipeline and task to the catalog to push disk images via pulp.

### DIFF
--- a/internal-services/catalog/pulp-push-disk-images-task.yaml
+++ b/internal-services/catalog/pulp-push-disk-images-task.yaml
@@ -1,0 +1,202 @@
+---
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: pulp-push-disk-images
+  labels:
+    app.kubernetes.io/version: "0.1.0"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: release
+spec:
+  description: >-
+    Tekton task to push disk images with pulp
+  params:
+    - name: snapshot_json
+      type: string
+      description: String containing a JSON representation of the snapshot spec
+    - name: concurrentLimit
+      type: string
+      description: The maximum number of images to be pulled at once
+      default: 3
+    - name: exodusGwSecret
+      type: string
+      description: Env specific secret containing the Exodus Gateway configs
+    - name: exodusGwEnv
+      type: string
+      description: Environment to use in the Exodus Gateway. Options are [live, pre]
+    - name: pulpSecret
+      type: string
+      description: Env specific secret containing the rhsm-pulp credentials
+    - name: udcacheSecret
+      type: string
+      description: Env specific secret containing the udcache credentials
+  results:
+    - name: result
+      description: Success if the task succeeds, the error otherwise
+  steps:
+    - name: pull-and-push-images
+      image: quay.io/konflux-ci/release-service-utils:cff1fd1e9d8fabc256ff17f6731b2c037a0a1102
+      env:
+        - name: EXODUS_CERT
+          valueFrom:
+            secretKeyRef:
+              name: $(params.exodusGwSecret)
+              key: cert
+        - name: EXODUS_KEY
+          valueFrom:
+            secretKeyRef:
+              name: $(params.exodusGwSecret)
+              key: key
+        - name: EXODUS_URL
+          valueFrom:
+            secretKeyRef:
+              name: $(params.exodusGwSecret)
+              key: url
+        - name: PULP_URL
+          valueFrom:
+            secretKeyRef:
+              name: $(params.pulpSecret)
+              key: pulp_url
+        - name: PULP_CERT
+          valueFrom:
+            secretKeyRef:
+              name: $(params.pulpSecret)
+              key: konflux-release-rhsm-pulp.crt
+        - name: PULP_KEY
+          valueFrom:
+            secretKeyRef:
+              name: $(params.pulpSecret)
+              key: konflux-release-rhsm-pulp.key
+        - name: UDC_URL
+          valueFrom:
+            secretKeyRef:
+              name: $(params.udcacheSecret)
+              key: url
+        - name: UDC_CERT
+          valueFrom:
+            secretKeyRef:
+              name: $(params.udcacheSecret)
+              key: cert
+        - name: UDC_KEY
+          valueFrom:
+            secretKeyRef:
+              name: $(params.udcacheSecret)
+              key: key
+        - name: "SNAPSHOT_JSON"
+          value: "$(params.snapshot_json)"
+      script: |
+        #!/usr/bin/env bash
+        set -ex
+
+        STDERR_FILE=/tmp/stderr.txt
+
+        exitfunc() {
+            local err=$1
+            local line=$2
+            local command="$3"
+            if [ "$err" -eq 0 ] ; then
+                echo -n "Success" > "$(results.result.path)"
+            else
+                echo "$0: ERROR '$command' failed at line $line - exited with status $err" \
+                  > "$(results.result.path)"
+                if [ -f "$STDERR_FILE" ] ; then
+                    cat "$STDERR_FILE" | tail -n 20 >> "$(results.result.path)"
+                fi
+            fi
+            exit 0 # exit the script cleanly as there is no point in proceeding past an error or exit call
+        }
+        # due to set -e, this catches all EXIT and ERR calls and the task should never fail with nonzero exit code
+        trap 'exitfunc $? $LINENO "$BASH_COMMAND"' EXIT
+
+        # Setup required variables
+        export EXODUS_GW_CERT=/tmp/exodus.crt
+        export EXODUS_GW_KEY=/tmp/exodus.key
+        export PULP_CERT_FILE=/tmp/pulp.crt
+        export PULP_KEY_FILE=/tmp/pulp.key
+        export UDCACHE_CERT=/tmp/udc.crt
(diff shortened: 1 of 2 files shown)


johnbieren:
Opening this now to get early feedback. The udcache variables are not complete but otherwise it should be good to review @davidmogar @theflockers @mmalina @happybhati

internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -0,0 +1,191 @@
...
+            # so create a custom auth file with just one entry
+            AUTH_FILE=$(mktemp)
+            select-oci-auth "${PULLSPEC}" > "$AUTH_FILE"
+            oras pull --registry-config "$AUTH_FILE" "$PULLSPEC"
+            NUM_MAPPED_FILES=$(jq '.staged.files | length' <<< ${COMPONENT})
+            for ((i = 0; i < $NUM_MAPPED_FILES; i++)) ; do
+                FILE=$(jq -c --arg i "$i" '.staged.files[$i|tonumber]' <<< $COMPONENT)
+                SOURCE=$(jq -r '.source' <<< $FILE)
+                FILENAME=$(jq -r '.filename' <<< $FILE)
+                gzip -d "${SOURCE}.gz" || echo "didn't find mapped file: ${SOURCE}.gz"
```
mmalina, line 146: Shouldn't this result in an error? I am not sure right how the `staged.files` array is generated. But surely all the files should be in the oci artifact, no?

internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -0,0 +1,191 @@
...
+                DESTINATION_FILE="${DESTINATION}/${FILENAME}"
+                # Albeit a very small one, this is a race condition since this is run in parallel
```
mmalina, line 148: I'm not sure what this means. What is a race condition? From what I can see, there is a very small chance that due to a race condition you might not catch this - if two running instances of the function are gonna use the same `DESTINATION_FILE`, in the moment that they check, none will exist yet, so both pass this check and eventually one will override the other. Is that what you meant? If we're concerned about this, we could do it another way - we could construct the list of all the destination filenames and save it in a file, one line per item. And then check if there are any duplicates: `test "$(cat desination_filenames.txt | wc -l)" = "$(cat desination_filenames.txt | sort | uniq | wc -l)"`. Or something like that. But maybe it's not really such a concern, right?

Again, I'm not sure where `staged` is generated. But the check for duplicates could be done there.

internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -0,0 +1,191 @@
...
+
+        # Change to the subdir with the images
+        cd ${DISK_IMAGE_DIR}
+
+        STAGED_JSON='{"header":{"version": "0.2"},"payload":{"files":[]}}'
+
+        # Add the files to the payload
+        for file in $(find * -type f); do
+            STAGED_JSON=$(jq --arg filename $(basename $file) --arg path "$file" \
+              '.payload.files[.payload.files | length] +=
```
mmalina, line 184: It's a small detail, but I think the `+` in `+=` is redundant here, because you're assigning this to a new array index and it didn't exist previously, so there is nothing really to add to.

Another option would be to use `+=` on the array level and wrap the new item in `[]` - this way you don't need to specify the new array index:
```sh
STAGED_JSON=$(jq --arg filename $(basename $file) --arg path "$file" \
    '.payload.files += 
    [{"filename": $filename, "relative_path": $path}]' <<< $STAGED_JSON)
```

internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -0,0 +1,191 @@
...
+            wait -n
+        done
+
+        # Change to the subdir with the images
+        cd ${DISK_IMAGE_DIR}
+
+        STAGED_JSON='{"header":{"version": "0.2"},"payload":{"files":[]}}'
+
+        # Add the files to the payload
+        for file in $(find * -type f); do
```
mmalina, line 182: Are we sure there are no spaces in the paths anywhere? I guess probably there shouldn't be?

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: https://github.com/konflux-ci/release-service-catalog/pull/463#discussion_r1661627128
Pretty sure I talked to David about it on slack

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: What you describe is the race condition, yes.
What do you mean not sure where staged is generated? staged.yaml?

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: doesn't seem to work
```
$ STAGED_JSON='{"header":{"version": "0.2"},"payload":{"files":[]}}'
$ for file in $(find * -type f) ; do   STAGED_JSON=$(jq --arg filename $(basename $file) --arg path "$file" \
    '.payload.files[.payload.files | length] = [{"filename": $filename, "relative_path": $path}]' <<< $STAGED_JSON); done
$ echo $STAGED_JSON | yq -P -I4
header:
    version: "0.2"
payload:
    files:
        - - filename: a
            relative_path: 1/FILES/a
        - - filename: b
            relative_path: 1/FILES/b
        - - filename: a
            relative_path: 2/FILES/a
        - - filename: b
            relative_path: 2/FILES/b
```

```
$ STAGED_JSON='{"header":{"version": "0.2"},"payload":{"files":[]}}'
$ for file in $(find * -type f) ; do   STAGED_JSON=$(jq --arg filename $(basename $file) --arg path "$file" \
    '.payload.files[.payload.files | length] += {"filename": $filename, "relative_path": $path}' <<< $STAGED_JSON); done
$ echo $STAGED_JSON | yq -P -I4
header:
    version: "0.2"
payload:
    files:
        - filename: a
          relative_path: 1/FILES/a
        - filename: b
          relative_path: 1/FILES/b
        - filename: a
          relative_path: 2/FILES/a
        - filename: b
          relative_path: 2/FILES/b
```

internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina: Fair enough.

internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina: Yeah, so that comment is not very clear imo. "this is a race condition" - rather, in a rare case a race condition may prevent us from detecting a duplicate destination here.

> What do you mean not sure where staged is generated? staged.yaml?

I am not sure where the staged part of the snapshot is generated - here it's already included on the input. But it must be generated somewhere beforehand. Or is this from the mapping? So in the RPA? So maybe a check should be added to gitlab ci in the repo to find duplicates there?

internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina: Yeah, `$SOURCE` should be `"$SOURCE"` - I'm pretty sure the shell check would tell you that once enabled :)

johnbieren: I added quotes to `$SOURCE`. Is there something else to change here?

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: Updated the comment. The mapping yes, in the RPA. I think a check in gitlab ci is something we should eventually do, yes. It also is kind of obvious to the eye if two components use the exact same values

johnbieren: @davidmogar should I create a story for the ci check Martin is mentioning? It is kind of unclear to me if this is even end code or just a hacky POC to get something out the door

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: Changed to your first suggestion

internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina: It depends. I was trying to point out that this is fragile because of possible spaces and you essentially said that it would break even before this. That is fixed, but that doesn't change the fragility of this. And now I can see that shellcheck doesn't like it either.

I would be fine with it if you're confident there won't be spaces. But now that we have shellcheck, you should probably change it, or you'll have to find a way to tell shellcheck to ignore it.

But I guess you should probably use a while loop as shellcheck suggests (including `find ... -print0`).

johnbieren: fixed

mmalina (dismissed):

johnbieren:
I need to merge this for testing that has to be done, so feel free to add feedback but I will fix it in a separate PR

internal-services/catalog/pulp-push-disk-images-task.yaml
davidmogar: That's ok to me.

You can also get hacky and use `mv -nv "$SOURCE" "${DESTINATION_FILE}"` and see if the output of that command contains `not overwritten`, failing if so... but at that point some of the mv commands would have succeeded.

PR #147, opened 2024-07-24
fix: only unzip files if they are zipped

johnbieren:


internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -162,7 +164,7 @@ spec:
                     echo " and filename value: $FILENAME. Failing..." >&2
                     exit 1
                 fi
-                mv "$SOURCE" "${DESTINATION_FILE}"
+                mv "$SOURCE" "${DESTINATION_FILE}" || echo "didn't find mapped file: ${SOURCE}"
```
mmalina, line 167: Previously the task would fail here if the file was missing. Now it will just skip it. Is that intentional? Probably yes, because we never want the internal services pipeline to fail if at all possible, right? But shouldn't we maybe indicate it in the result that it failed?

internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -152,7 +152,9 @@ spec:
                 FILE=$(jq -c --arg i "$i" '.staged.files[$i|tonumber]' <<< "$COMPONENT")
                 SOURCE=$(jq -er '.source' <<< "$FILE")
                 FILENAME=$(jq -er '.filename' <<< "$FILE")
-                gzip -d "${SOURCE}.gz" || echo "didn't find mapped file: ${SOURCE}.gz"
+                if [ -f "${SOURCE}.gz" ] ; then
```
mmalina, line 155: I agree this is nicer. But just to be clear, the behavior stayed the same here, right? (unless the gz file is corrupted or something)

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: It still wouldn't "fail" (the trap would catch it and it would exit 0). Execution would stop though. I am a little confused as to why this is though (my run stopped here because my mapping was wrong)
```
$ ls ababba | echo "didn't find"
didn't find
ls: cannot access 'ababba': No such file or directory
$ echo $?
0
```

Shouldn't the `|| echo` make the return code still be 0?
Anyways, the context for this is the `.qcow2` files are not zipped, the others are

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: Mentioned in other comment but I only just now realized that the qcow2s are not zipped

internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina: > Anyways, the context for this is the .qcow2 files are not zipped, the others are

I think that would be a useful comment above that check.

In your example above, you use `|` instead of `||`. And that will depend on whether you have `set -u` or not. But yeah `|| echo` should make it pass always.

mmalina: It works for me:
```sh
#!/usr/bin/env bash
set -eu

exitfunc() {
    local err=$1
    local line=$2
    local command="$3"
    if [ "$err" -eq 0 ] ; then
        echo "Success"
    else
        echo "$0: ERROR '$command' failed at line $line - exited with status $err"
    fi
    exit 0 # exit the script cleanly as there is no point in proceeding past an error or exit call
}

trap 'exitfunc $? $LINENO "$BASH_COMMAND"' EXIT


ls asfasda || echo blabla

echo something else
```
```sh
$ ./or.sh
ls: asfasda: No such file or directory
blabla
something else
Success
```

johnbieren: Comment added. Going to leave it as not failing if the echo hits. If Ralph/Scott ask for it to be changed later on, we can. I had asked them previously, didn't get an answer, and then talked to David on slack about it and went with not failing

internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina: Yeah, ok. Would mind adding a comment above this?

mmalina: Comment added, thanks.

internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina: ok, thanks

mmalina (approved):

internal-services/catalog/pulp-push-disk-images-task.yaml
johnbieren: Sweet, that's the behavior I want and is the one I think should happen now

PR #149, opened 2024-07-30
fix: manually create docker config for pulp

johnbieren:
This commit modifies the pulp-push-to-cdn-task to manually create the ~/.docker/config.json file instead of relying on tekton to do it via volume mounts as changing tekton config flags will make tekton no longer add the volume mount. So, let's not depend on it.


davidmogar (approved):

happybhati (approved):
/lgtm

PR #154, opened 2024-08-14
feat(RELEASE-1108): push disk images to cgw

scoheb:
- WIP

### DIFF
--- a/internal-services/catalog/pulp-push-disk-images-task.yaml
+++ b/internal-services/catalog/pulp-push-disk-images-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: pulp-push-disk-images
   labels:
-    app.kubernetes.io/version: "0.1.0"
+    app.kubernetes.io/version: "0.2.0"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
@@ -31,6 +31,24 @@ spec:
     - name: udcacheSecret
       type: string
       description: Env specific secret containing the udcache credentials
+    - name: cgwHostname
+      type: string
+      description: Env specific hostname for content gateway
+    - name: cgwSecret
+      type: string
+      description: Env specific secret containing the content gateway credentials
+    - name: productName
+      type: string
+      description: product name
+    - name: productCode
+      type: string
+      description: product code
+    - name: productVersionName
+      type: string
+      description: product version name
+    - name: component_json
+      type: string
+      description: String containing a JSON representation of the components listed in the data.mapping in RPA
   results:
     - name: result
       description: Success if the task succeeds, the error otherwise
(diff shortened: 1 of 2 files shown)


internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -211,3 +241,131 @@ spec:
...
+        #    data = json.load(file)
+
+        os.makedirs(WORKSPACE_DIR, exist_ok=True)
+
+        productName = "$(params.productName)" #data['contentGateway']['productName']
+        productCode = "$(params.productCode)" #data['contentGateway']['productCode']
+        productVersionName = "$(params.productVersionName)" #data['contentGateway']['productVersionName']
+        components = json.loads("$(params.components_json)") #data['contentGateway']['components']
+
+        content_list = os.listdir(CONTENT_DIR)
```
johnbieren, line 280: where is CONTENT_DIR defined?

internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -31,9 +31,29 @@ spec:
...
+    - name: productVersionName
+      type: string
+      description: product version name
+    - name: component_json
+      type: string
+      description: String containing a JSON representation of the components listed in the data.mapping in RPA
   results:
     - name: result
       description: Success if the task succeeds, the error otherwise
+    - name: download-dir
```
johnbieren, line 55: This doesn't seem used, and I don't really see what the value of it could be. Why would we want to say a dir name in a pod as a result of the InternalRequest pipelineRun and pass that up to the managed pipelineRun which isn't even on the same cluster? I am assuming that is why we'd want the result because so far the only use of IR plr results is to return the data to the managed plr

scoheb: removed

internal-services/catalog/pulp-push-disk-images-task.yaml
scoheb: good catch. fixed

internal-services/catalog/pulp-push-disk-images-task.yaml
```diff
@@ -211,3 +239,120 @@ spec:
...
+        }
+
+        def generate_download_url(file_name):
+            """
+            Generate a download URL in this format:
+            /content/origin/files/sha256/{checksum[:2]}{checksum}/{file_name}
+            """
+            prefix = "/content/origin/files/sha256"
+            sha256_hash = hashlib.sha256()
+            with open(CONTENT_DIR + "/" + file_name, "rb") as f:
```
johnbieren, line 289: CONTENT_DIR is here too but I don't see it defined anywhere

scoheb: fixed

johnbieren (approved):

PR #157, opened 2024-08-15
fix(KFLUXBUGS-1278): make iib errors available to users

johnbieren:
This commit exposes the iib error if one occurred in the iib operation and bubbles it up to the pipelineRun as a result.


johnbieren:
@hacbs-release/release-team PTAL. I haven't tested it yet but will report back once I do

johnbieren:
Okay it is tested now

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -226,8 +230,15 @@ spec:
             # remove the history as it breaks the results build up
             jq -r 'del(.state_history)' <<< \${build_info} | jq -c . > $(results.jsonBuildInfo.path)
             case \${state} in
-              "complete") break ;;
-              "failed") break ;;
+              "complete") echo -n "Build succeeded so IIB log is not included" > \$(results.iibLog.path); break ;;
```
mmalina, line 233: Even for a successful build it would be nice to pass the iib build url and print it in the release pipeline, no?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina, line 1: You will have to sync this with a corresponding release-service-catalog change, right?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: Not sure - would anyone really care to check a log in case of success? I never look at any logs personally when things pass, that is why I did it this way

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: ok, cool.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: OK, if you don't want to add it, it's fine, it's not required. But I think it's a nice thing to have - if something unexpected happens and you want to debug (as a user). And in this case it would be trivial to do. But it's true that we don't have anything like that for other pipelines here.

mmalina (approved):
LGTM. But once again, this shouldn't be merged until this is also addressed in the release-service-catalog repo - otherwise we might mistakingly consider all iib runs successful, even if they fail. At least I assume that would happen - I haven't checked.

johnbieren:
> LGTM. But once again, this shouldn't be merged until this is also addressed in the release-service-catalog repo - otherwise we might mistakingly consider all iib runs successful, even if they fail. At least I assume that would happen - I haven't checked.

Good point. So really I shouldn't merge this until the catalog change is merged and promoted to staging (and then merge this PR promotion to stable branch when that change goes to production in catalog)

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: I am not sure the catalog task will pass e2e until this is merged actually because I exit the catalog task with the exitCode result. So I may need to merge it in stage to get e2e to pass and just have stage always pass for a week

johnbieren: I'll probably merge this once the catalog PR is approved and e2e is fixed. Then I won't promote this to prod until the catalog task change is being promoted to prod

theflockers:
@happybhati your work depends on this one. Your task should be easier once this is merged.

theflockers (approved):
looks good!

johnbieren:
> @happybhati your work depends on this one. Your task should be easier once this is merged.

If https://github.com/konflux-ci/release-service-catalog/pull/516 gets approved, I will merge both that and this one tomorrow

PR #164, opened 2024-08-19
fix: update image used in pulp task

mmalina:
The updated image logs stuff in pulp_push_wrapper.py to stdout instead of stderr. This will help with debugging.

Here's the related change: https://github.com/konflux-ci/release-service-utils/pull/225


scoheb:
#163 takes care of this...

PR #171, opened 2024-08-28
fix: decrease iib log size

johnbieren:
The tekton result was still over the max, so cut the log size further.


internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -232,8 +232,8 @@ spec:
             url=\$(jq -r ".logs.url" <<< \${build_info})
             echo IIB log url is: "\${url}" > \$(results.iibLog.path)
             echo Printing last 2,000 characters from the log >> \$(results.iibLog.path)
```
scoheb, line 234: this seems wrong

johnbieren: Thanks, fixed. Missed it because of the `,`

theflockers (approved):
Lets try this one. Eventually we might need to count the chars instead.

johnbieren:
> Lets try this one. Eventually we might need to count the chars instead.

I am counting the chars
I use `tail -c` not `tail -n`

PR #173, opened 2024-08-28
fix(KFLUXBUGS-1278): make iib errors available to users

johnbieren:
This commit exposes the iib error if one occurred in the iib operation and bubbles it up to the pipelineRun as a result. It also changes the iib-add-fbc-fragment-to-index-image task to never fail.


theflockers (commented):
Lgtm once the shell check warning goes away.

theflockers (approved):
.

PR #174, opened 2024-08-29
fix: indexImageDigests should exist in all cases

theflockers:
this commit fixes the indexImageDigests result creation for all cases due to that in case of success, its creation was conditioned to a non-empty parameter in the json response


theflockers:
fixed by another PR.

PR #176, opened 2024-08-30
fix: increase timeout for push-disk-images IR

mmalina:
The disk images for RHEL AI are very big and
we repeatedly saw timeouts. We will make it
customizable here:
https://issues.redhat.com/browse/RELEASE-1143

But for now, let's increase the value.


scoheb (approved):

scoheb:
Reminder this needs to get into the stable branch for production

mmalina:
OK, here's one for stable: https://github.com/hacbs-release/app-interface-deployments/pull/177

PR #177, opened 2024-08-30
fix: increase timeout for push-disk-images IR

mmalina:
The disk images for RHEL AI are very big and
we repeatedly saw timeouts. We will make it
customizable here:
https://issues.redhat.com/browse/RELEASE-1143

But for now, let's increase the value.


scoheb (approved):

johnbieren:
The commit isn't signed

mmalina:
> The commit isn't signed

Ah, that's probably because I cherry-picked from the other PR/branch. Fixed, thanks.

johnbieren (approved):

PR #182, opened 2024-09-02
refactor(RELEASE-1024): update checkton version and allow it to fail

mmalina:
Originally, our goal was to write a special
script to parse the sarif file and print a nice summary.

But since then, Adam improved the output of the checkton action, so that it has all we need (including link to wiki for each issue), so we can just use that.

This the same as was done in the main catalog repo here: https://github.com/konflux-ci/release-service-catalog/pull/536

I tested it here by modifying a task and then it ran as expected (with 1 issue): https://github.com/hacbs-release/app-interface-deployments/actions/runs/10666209886/job/29561166377?pr=182

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -11,7 +11,7 @@ jobs:
   yamllint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


seanconroy2021 (approved):
lgtm

PR #184, opened 2024-09-05
chore(KFLUXBUGS-1278): promote main to stable

johnbieren:
Promote the iib pipeline and task from main to stable so that IIB errors are better exposed to users.

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -1,10 +1,10 @@
 ---
-apiVersion: tekton.dev/v1beta1
+apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.2.0"
+    app.kubernetes.io/version: "0.2.1"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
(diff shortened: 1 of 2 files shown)


johnbieren:
This should be merged Wed, Sept 11 when the catalog promotion to production happens

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina, line 1: Just to be clear, you're promoting that one change, right? So why not cherry-pick that commit? Or was the current one in stable more outdated than just the one change? It seems weird to have "promote branch" commits in the stable branch instead of the real commits, but I guess it's probably acceptable since we're moving away from this repo soon anyway.

johnbieren: Because main and stable are so far diverged (probably due to many commits just like this one). But yes, just that one change

mmalina: So maybe at least modify the commit message to be the same as the original?

theflockers: I would say to synchronize these branches, but as we will move this content to the catalog, we can keep doing this way until the move happens.

theflockers (approved):

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: I updated the commit message

PR #190, opened 2024-10-01
fix: enable more logging for kinit

mmalina:
Recently, we had issues when doing kinit, it would say:

kinit: Generic error (see e-text) while
getting initial credentials

It was suggested to us to enable
kerberos trace logging so that next time
we have more details to report.

Discussion: https://redhat-internal.slack.com/archives/C04J6G6F14H/p1727717957150979


internal-services/catalog/check-embargoed-cves-task.yaml
```diff
@@ -69,6 +69,7 @@ spec:
           # workaround kinit: Invalid UID in persistent keyring name while getting default ccache
           export KRB5CCNAME=`mktemp`
           export KRB5_CONFIG=`mktemp`
+          export KRB5_TRACE=/dev/stdout
```
lslebodn, line 72: I have a feeling it will break passing stdout to jq :-)
`curl --retry 3 --negotiate -u : ${OSIDB_URL}/auth/token | jq -r '.access'`

similar on other places

mmalina: Damn. Good point, thanks. I didn't realize curl would also print tons of kerberos logs as well. So probably just use `/dev/stderr` instead? Fixed.

lslebodn: `/dev/stderr` should be fine

mmalina: Thanks.

PR #192, opened 2024-10-01
feat: add sign-image pipeline

mmalina:
This pipeline yaml is a copy of the pipeline we currently use: hacbs-signing-pipeline

The goal here is to enable retry in the request-signature task, see here:
https://issues.redhat.com/browse/RELEASE-1194

For that, I needed to update the task reference. But that broke the task because the current version of the task does not have the signature_data result.

In order to test this, I am deploying this modified pipeline under a different name.

Once this is tested, we may actually switch to using this going forward - we can evaluate this afterwards.

### DIFF
--- a/internal-services/catalog/sign-image-pipeline.yaml
+++ b/internal-services/catalog/sign-image-pipeline.yaml
@@ -1,0 +1,106 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: sign-image
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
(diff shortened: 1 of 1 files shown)


scoheb (approved):

johnbieren:
We need a migration ticket for this like https://issues.redhat.com/browse/RELEASE-1040
The goal is to have less tekton resources in this repo (0 actually)

mmalina:
> We need a migration ticket for this like https://issues.redhat.com/browse/RELEASE-1040 The goal is to have less tekton resources in this repo (0 actually)

I created https://issues.redhat.com/browse/RELEASE-1206 and also https://issues.redhat.com/browse/RELEASE-1207 to actually start using it.

PR #195, opened 2024-10-11
fix(RELEASE-1101): linting issues in pulp-push-disk-images task

happybhati:


internal-services/catalog/pulp-push-disk-images-task.yaml
mmalina, line 1: Can you bump the task version?

johnbieren (commented):
Did you test the entire file through the linter? I am surprised it didn't complain about things like line 179 where the variable isn't quoted (which I have seen it complain about about 90 other times)

internal-services/catalog/pulp-push-disk-images-task.yaml
happybhati: Done

happybhati:
> Did you test the entire file through the linter? I am surprised it didn't complain about things like line 179 where the variable isn't quoted (which I have seen it complain about about 90 other times)

I am unsure why it did not complain, but I tested the entire file.

johnbieren (approved):

mmalina:
> Did you test the entire file through the linter? I am surprised it didn't complain about things like line 179 where the variable isn't quoted (which I have seen it complain about about 90 other times)

You mean stuff like this?
```sh
FILENAME=$(jq -er '.filename' <<< "$FILE")
```
I noticed this before - if you're assigning it to a variable, it will not require "". But I'm actually not sure why.

This will not work:
```sh
$ FILENAME=my name with spaces
bash: name: command not found
```

But this will work:
```sh
$ FILENAME=$(echo my name with spaces)
$ echo $FILENAME
my name with spaces
```

Magic?

johnbieren:
> > Did you test the entire file through the linter? I am surprised it didn't complain about things like line 179 where the variable isn't quoted (which I have seen it complain about about 90 other times)
> 
> You mean stuff like this?
> 
> ```shell
> FILENAME=$(jq -er '.filename' <<< "$FILE")
> ```
> 
> I noticed this before - if you're assigning it to a variable, it will not require "". But I'm actually not sure why.
> 
> This will not work:
> 
> ```shell
> $ FILENAME=my name with spaces
> bash: name: command not found
> ```
> 
> But this will work:
> 
> ```shell
> $ FILENAME=$(echo my name with spaces)
> $ echo $FILENAME
> my name with spaces
> ```
> 
> Magic?

Oh interesting. Nice investigation

mmalina:
BTW, I found out why - I commented in another PR today, but forgot where that was. The reason is that world spliting (and wildcard expansion and maybe something else) does not happen when assigning variables.

PR #196, opened 2024-10-11
fix(RELEASE-1100): linting issues in publish-index-image-task

happybhati:

PR #197, opened 2024-10-15
fix: use virtual environment for gitlint

happybhati:
This PR fixes the `gitlint` installation issue in the CI pipeline caused by PEP 668. A virtual environment is now used to install `gitlint`, avoiding conflicts with the system-managed Python environment.
See [comment](https://github.com/hacbs-release/app-interface-deployments/pull/196#issuecomment-2414224881) for more details.


happybhati:
Fixed by https://github.com/hacbs-release/app-interface-deployments/pull/198

PR #198, opened 2024-10-16
fix: gitlint installation

mmalina:
Latest ubuntu does not allow system wide pip install. It can be overriden with `--break-system-packages`, but it's better to create a virtual env.


happybhati (approved):

johnbieren:
I'd think we'd need a PR like this in all of our repos, no?

johnbieren:
Thank you!

PR #206, opened 2024-10-23
feat(RELEASE-1191): use prod radas for signing

johnbieren:
Based on https://github.com/hacbs-release/app-interface-deployments/blob/main/internal-services/catalog/sign-image-pipeline.yaml#L62, it seems the UMB_ and SSL_ variables relate to the UMB authentication, so I removed the staging values from those in the configmaps that were using the stage umb url


internal-services/config/signing/production/hacbs-signing-pipeline-config-staging-redhatrelease2.yaml
```diff
@@ -13,9 +13,9 @@ data:
   SIG_KEY_ID: "4096R/37036783 SHA-256"
   SIG_KEY_NAME: "redhate2etesting"
   SSL_CERT_FILE_NAME: "hacbs-signing-pipeline.pem"
-  SSL_CERT_SECRET_NAME: "hacbs-signing-pipeline-staging-certs"
+  SSL_CERT_SECRET_NAME: "hacbs-signing-pipeline-certs"
```
scoheb, line 16: I do not think these certs work with PYXIS_URL: "https://pyxis.stage.engineering.redhat.com"

scoheb (commented):
Keep in mind for RELEASE-1192, we also have these https://gitlab.cee.redhat.com/releng/konflux-release-data/-/tree/main/tenants-config/cluster/stone-prod-p02/managed/rhtap-releng-tenant/cm?ref_type=heads

internal-services/config/signing/production/hacbs-signing-pipeline-config-staging-redhatrelease2.yaml
johnbieren: This line https://github.com/hacbs-release/app-interface-deployments/blob/main/internal-services/catalog/sign-image-pipeline.yaml#L62 says that the SSL_CERT_SECRET_NAME is for umb, that is why I guess it should be updated. So this is actually for pyxis? Happy to update if I guessed wrong

johnbieren:
> Keep in mind for RELEASE-1192, we also have these https://gitlab.cee.redhat.com/releng/konflux-release-data/-/tree/main/tenants-config/cluster/stone-prod-p02/managed/rhtap-releng-tenant/cm?ref_type=heads

Thanks, did not know about those. I will update those the same way I do these once it is decided how these should be updated

internal-services/config/signing/production/hacbs-signing-pipeline-config-staging-redhatrelease2.yaml
johnbieren: @mmalina you don't know the answer to this do you?

mmalina: Hmm, it's weird, but AFAICS, the same key+cert is used for both umb and pyxis? If that's the case, then you will need to modify the secrets to contain separate key+cert values for umb and pyxis, because umb will always be prod, but pyxis will be prod vs stage.

Here's why I think the same one is used for both:

The set-env task is here: https://gitlab.cee.redhat.com/signing/hacbs-signing-pipeline/-/blob/main/ansible/roles/hacbs-signing-pipeline/templates/openshift/templates/signing-pipeline-stage-template.yml?ref_type=heads

It returns:

- ssl_cert_secret_name
- ssl_key_file_name
- ssl_cert_file_name

All of them taken from the secret passed to the pipeline.

[request-signature](https://github.com/mmalina/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/request-signature.yml) uses these to set two env vars: UMB_CERT_PATH, UMB_KEY_PATH

- the umb env vars are read here: https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/operator-pipeline-images/operatorcert/umb.py#L113

[upload-signature](https://github.com/mmalina/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/upload-signature.yml) uses these to set two env vars: PYXIS_CERT_PATH, PYXIS_KEY_PATH

mmalina: I pinged Ales Raszka to take a look. He should be back next week (Monday is public holiday in CZ, so on Tuesday).

johnbieren: Thanks Martin, I'll wait for that to make any adjustments

Allda: Hey, @mmalina is right. The pipeline uses a single cert for both Pyxis and UMB authentication. If you have multiple certs you will need update the pipeline to server your needs. Other alternative is to request a Pyxis and UMB permission for a single service account represented by the one cert.

johnbieren: @mmalina is there a ticket for switching to use our copy of the pipeline (maybe it is already done)? It seems like this would be blocked on that then

mmalina: Here's the Jira: https://issues.redhat.com/browse/RELEASE-1207

But we had an open question about whether or not to also have our own copy of the tasks, not just the pipeline. I think this is a good enough reason to also have a copy of the tasks so that we can modify them to support different credentials for pyxis and umb. I am adding a comment there.

mmalina: Update: The request-signature and upload-signature actually each need just one secret each (to access umb and pyxis, respectively). So we would only need to modify the set-env task.

johnbieren: Thanks!

johnbieren:
@mmalina I think with the app-sre change merged this is ready for another review

mmalina (approved):

mmalina:
> @mmalina I think with the app-sre change merged this is ready for another review

Yeah. It was not clear what the staging vs. production folders are for, since we also have main vs. stable branches. But Johnny found out:

> prod [https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/stoneso[…]ervices/cicd/stonesoup-internal-services.appsrep09ue1.yaml](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/stonesoup/internal-services/cicd/stonesoup-internal-services.appsrep09ue1.yaml#L120) staging [https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/stoneso[…]ervices/cicd/stonesoup-internal-services.appsres09ue1.yaml](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/stonesoup/internal-services/cicd/stonesoup-internal-services.appsres09ue1.yaml#L120)
[16:53](https://redhat-internal.slack.com/archives/D04MF7AP8NR/p1732722800592459)
so we use staging dir with main branch on stage and production dir with stable branch on prod

So it turns out we do use staging cert for staging signing in prod cluster. So it's still not clear if this will work - it's possible that stage Pyxis will not work with the prod cert. But let's see.

internal-services/config/signing/production/hacbs-signing-pipeline-config-staging-redhatrelease2.yaml
johnbieren: @Allda is it possible for you to check if our prod pyxis account has perms for stage pyxis?

mmalina:
> This PR was updated to now modify all configMaps with staging in their name or path to use prod UMB credentials so we will use prod radas. Moving it back to ready for review

IOW, all occurrences of `umb.stage.api.redhat.com` in the whole repo were changed to `umb.api.redhat.com` and `hacbs-signing-pipeline-nonprod` to `hacbs-signing-pipeline`. Right?

mmalina (approved):

johnbieren:
> > This PR was updated to now modify all configMaps with staging in their name or path to use prod UMB credentials so we will use prod radas. Moving it back to ready for review
> 
> IOW, all occurrences of `umb.stage.api.redhat.com` in the whole repo were changed to `umb.api.redhat.com` and `hacbs-signing-pipeline-nonprod` to `hacbs-signing-pipeline`. Right?

Yes, but also the `UMB_SSL_CERT_SECRET_NAME` values were all changed to the prod one, but the pyxis ones left alone

mmalina:
> > > This PR was updated to now modify all configMaps with staging in their name or path to use prod UMB credentials so we will use prod radas. Moving it back to ready for review
> > 
> > 
> > IOW, all occurrences of `umb.stage.api.redhat.com` in the whole repo were changed to `umb.api.redhat.com` and `hacbs-signing-pipeline-nonprod` to `hacbs-signing-pipeline`. Right?
> 
> Yes, but also the `UMB_SSL_CERT_SECRET_NAME` values were all changed to the prod one, but the pyxis ones left alone

Yeah, good point.

mmalina (approved):

PR #207, opened 2024-10-24
fix: empty variable should not keep quotes (#201)

happybhati:
the PR fixes the bug caused by when an empty AUTH_PARAM is passed the quotes are preserved, causing skopeo to complain on the unknown parameter.


mmalina (approved):

mmalina:
It turns out that before this fix the task was completely broken even if you were using registry proxy as source repo (and so the condition was evaluated as false). The reason is that in that case `${AUTH_PARAM}` would be undefined and we would pass an extra `""` param which would break the skopeo call. Now that we use `"${AUTH_PARAM[@]}"` it should be fine - I checked and if the array is undefined, you end up with no extra param.

PR #208, opened 2024-10-31
fix(RELEASE-1271): use bundle resolver in signing pipeline

mmalina:
The old bundle reference was deprecated in Tekton 1.15 and removed in 1.16 and app-sre just deployed 1.16 which broke signing in Konflux. Updating to bundle resolver format should fix it.

Note that this pipeline definition is still not really used, we use the hacbs-signing-pipeline from a different repo, but this is a copy we have and it's useful
for testing.

### DIFF
--- a/internal-services/catalog/sign-image-pipeline.yaml
+++ b/internal-services/catalog/sign-image-pipeline.yaml
@@ -41,9 +41,16 @@ spec:
 
     - name: request-signature
       taskRef:
-        name: request-signature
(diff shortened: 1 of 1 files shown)


happybhati (dismissed):

happybhati:
Noticed a Linters / yamllint line too long error.

Just a note: we need similar change for stable as well.

happybhati (approved):

PR #213, opened 2024-11-01
feat(RELEASE-1223): validate created advisory against schema

johnbieren:
This commit modifies the create-advisory-task to validate the generated advisory.yaml against the advisory schema before pushing it to the repo.

### DIFF
--- a/internal-services/catalog/create-advisory-task.yaml
+++ b/internal-services/catalog/create-advisory-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: create-advisory-task
   labels:
-    app.kubernetes.io/version: "0.9.1"
+    app.kubernetes.io/version: "0.10.0"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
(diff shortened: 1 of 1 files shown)


happybhati (approved):

seanconroy2021 (approved):
LGTM

internal-services/catalog/create-advisory-task.yaml
```diff
@@ -157,6 +157,9 @@ spec:
           /home/utils/apply_template.py -o $ADVISORY_FILEPATH --data "$DATA" \
             --template /home/templates/advisory.yaml.jinja
 
+          # Ensure the created advisory file passes the advisory schema
+          check-jsonschema --schemafile schema/advisory.json "$ADVISORY_FILEPATH"
```
mmalina, line 161: How does this does? Where is the schema directory and file coming from? It's definitely not in our utils image: https://github.com/konflux-ci/release-service-utils/blob/main/Dockerfile

johnbieren: It's a relative path. It's in the advisory repo. Just like `ADVISORY_DIR="data/advisories/$(params.origin)/${YEAR}/${ADVISORY_NUM}"`

mmalina: Ah, I see, thanks.

PR #224, opened 2024-11-08
feat(RELEASE-1283): file-updates secret update

scoheb:
- allow file-updates-secret to be configurable

### DIFF
--- a/internal-services/catalog/file-updates-pipeline.yaml
+++ b/internal-services/catalog/file-updates-pipeline.yaml
@@ -29,6 +29,10 @@ spec:
     - name: application
       type: string
       description: Application being released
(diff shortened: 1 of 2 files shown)


scoheb:
forced merged since file-updates is not used by an project in konflux-release-data

johnbieren:
> forced merged since file-updates is not used by an project in konflux-release-data

??
run-file-updates is used in the rh-advisories pipeline. That is our most used pipeline of any

scoheb:
no team specifies `.spec.data.fileUpdates` in https://gitlab.cee.redhat.com/releng/konflux-release-data. therefore the task is never exercised.

PR #225, opened 2024-11-11
feat(RELEASE-1283): file-updates updates

scoheb:
* allow file-updates-secret to be configurable
* change reference to RHTAP to Konflux

### DIFF
--- a/internal-services/catalog/file-updates-pipeline.yaml
+++ b/internal-services/catalog/file-updates-pipeline.yaml
@@ -29,6 +29,10 @@ spec:
     - name: application
       type: string
       description: Application being released
+    - name: file_updates_secret
+      type: string
+      default: "file-updates-secret"
+      description: The credentials used to update the git repo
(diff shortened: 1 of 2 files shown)


happybhati (approved):

johnbieren:
Why do we need the secret to be configurable?

scoheb:
> Why do we need the secret to be configurable?

Right now, the task is generic in that it can update multiple repos. Each repo may require its own project access token so a single secret may not work. 

In addition, we may want to support creating MR for stage or prod and therefore we might to specify that via the RPA.

johnbieren (commented):
Got it. Can you bump the version things? And when the catalog PR is opened for this change (adding the params to the internal request) can you link it to this?

johnbieren (approved):
I guess the version strings weren't bumped in main and this is stable so doesn't make sense now since we would have a conflict

PR #227, opened 2024-11-18
chore(RELEASE-1040): remove file-updates

scoheb:
- migration performed in RELEASE-1283


internal-services/catalog/file-updates-pipeline.yaml
```diff
@@ -1,57 +0,0 @@
----
-apiVersion: tekton.dev/v1beta1
-kind: Pipeline
-metadata:
-  name: file-updates
```
johnbieren, line 5: I think we need to change the migrated version to this name before we remove this here

scoheb: done here: https://github.com/konflux-ci/release-service-catalog/pull/682/files#diff-50b7aa963fdf55a7ea15a3ad71d8cd3b3b8d314d07fea08bf2acbbba9ef3bc68R28

johnbieren (approved):
Approving, but please merge it only when 682 is being merged

johnbieren:
@scoheb we can merge this now, right?

scoheb:
Don't we need my catalog PR to land in staging?

johnbieren:
It actually is in staging since I promoted after it merged, but no, we remove from main when it goes to development and stable when it goes to production. See https://issues.redhat.com/browse/RELEASE-1211 for an explanation

PR #229, opened 2024-11-20
fix(KFLUXBUGS-1896): do not hardcode errata secret name for advisories

johnbieren:
This commit changes the create-advisory task and pipeline to accept a parameter for the errata secret instead of hardcoding it. This allows us to differentiate stage and prod.

### DIFF
--- a/internal-services/catalog/create-advisory-pipeline.yaml
+++ b/internal-services/catalog/create-advisory-pipeline.yaml
@@ -4,7 +4,7 @@ kind: Pipeline
 metadata:
   name: create-advisory
   labels:
-    app.kubernetes.io/version: "0.4"
+    app.kubernetes.io/version: "0.5"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: advisory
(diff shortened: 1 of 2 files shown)


internal-services/catalog/create-advisory-pipeline.yaml
```diff
@@ -31,6 +31,9 @@ spec:
     - name: advisory_secret_name
       type: string
       description: The name of the secret that contains the advisory creation metadata
+    - name: errata_secret_name
```
mmalina, line 34: Would it make sense to set a default here? Perhaps the stage one? That way you don't have to worry about syncing the change in the catalog repo. But maybe it's ok if you make sure to change catalog first. Just a thought.

johnbieren: If only this task had been migrated to the internal dir in the catalog repo!
I think it is better to be consistent and follow how advisory_secret_name works even if it means we have to merge both PRs around the same time. I personally don't want to add code that we are only adding because of a plumbing limitation

mmalina (approved):

happybhati (approved):

PR #230, opened 2024-11-21
fix(KFLUXBUGS-1896): do not hardcode errata secret name for advisories

johnbieren:
This commit changes the create-advisory task and pipeline to accept a parameter for the errata secret instead of hardcoding it. This allows us to differentiate stage and prod.


johnbieren:
To be merged with November 27th release-service-catalog production promotion

mmalina (approved):

PR #231, opened 2024-11-22
feat(RELEASE-1242): make add-fbc-contribution idempotent

theflockers:
This PR modifies the task add-fbc-fragment-to-index-image idempotent by checking IIB endpoint for builds for the current fbc_fragment, considering only prod builds and builds complete or in progress.

In addition, it also fixes shellcheck lint errors.

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.3.1"
+    app.kubernetes.io/version: "0.3.2"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
@@ -83,11 +83,16 @@ spec:
       script: |
         #!/usr/bin/env bash
         #
+        # all in line shellcheck disables will be handled in the PR below:
+        # https://github.com/hacbs-release/app-interface-deployments/pull/236
+
         isfbcOptIn() {
           TMPFILE=$(mktemp)
           PYXIS_URL="https://pyxis.engineering.redhat.com/v1"
 
+          # shellcheck disable=SC2162 # (info): read without -r will mangle backslashes.
           IFS="/" read REGISTRY REPO IMAGE <<< "${1}"
+          # shellcheck disable=SC2162 # (info): read without -r will mangle backslashes.
           IFS=":" read IMAGE TAG <<< "${IMAGE}"
 
           FETCH_URL="${PYXIS_URL}/repositories/registry/${REGISTRY}/repository/${REPO}/${IMAGE}/tag/${TAG}"
@@ -95,38 +100,41 @@ spec:
           # strips the last "/tag" in case $TAG is not set
           [ -z "${TAG}" ] && FETCH_URL=${FETCH_URL%/tag*}
 
+          # shellcheck disable=SC2086 # (info): Double quote to prevent globbing and word splitting.
           curl --negotiate -u: "${FETCH_URL}" -o $TMPFILE
+
           # prints "false" in case .fbc_opt_in entry is missing
+          # shellcheck disable=SC2086 # (info): Double quote to prevent globbing and word splitting.
           jq -e -r '.fbc_opt_in //false' $TMPFILE && rm -f $TMPFILE
+        }
+
+        # checks if there is any previous build for the same fbc_fragment.
+        # in case multiple builds are found, returns only the last one.
+        check_previous_build() {
+          user="${1}"
+          from_index="${2}"
+          fbc_fragment="${3}"
+
+          # fetch only builds in progress or completed
+          curl -s "${IIB_SERVICE_URL}/builds?user=${user}&from_index=${from_index}" | \
+            jq --arg fbc_fragment "${fbc_fragment}" \
+              '[.items[] |select(.fbc_fragment==$fbc_fragment and .state!="failed")][0] // empty'
         }
 
         # performs kerberos authentication.
         base64 -d /mnt/service-account-secret/keytab > /tmp/keytab
 
         KRB5_TEMP_CONF=$(mktemp)
+        KRB5_PRINCIPAL=$(cat /mnt/service-account-secret/principal)
+
         echo "${KRB5_CONF_CONTENT}" > "${KRB5_TEMP_CONF}"
         export KRB5_CONFIG="${KRB5_TEMP_CONF}"
+
         export KRB5_TRACE=/dev/stderr
 
-        /usr/bin/kinit -V $(cat /mnt/service-account-secret/principal) -k -t /tmp/keytab
+        /usr/bin/kinit -V "${KRB5_PRINCIPAL}" -k -t /tmp/keytab
 
         set -x
-        # check if this fbc fragment is opt-in
-        echo "Fetching the image bundle from $(params.fbcFragment)..."
-        PULL_SPEC_LIST=$(opm render $(params.fbcFragment) | jq -r \
-        'select(.schema == "olm.bundle") | "\(.image)" | split("@")[0]' |uniq)
-        for PULL_SPEC in ${PULL_SPEC_LIST}; do
-            # make sure they query is done using the internal name instead of the public
-            PULL_SPEC=$(sed  's/registry.redhat.io/registry.access.redhat.com/g' <<< $PULL_SPEC)
-
-            echo "Attempting to fetch from ${FETCH_URL} to check if fragment is \`fbc_opt_in==true\`..."
-            fbcOptIn+=(`isfbcOptIn ${PULL_SPEC}`)
-        done
-
-        fbcOptIn=($(printf "%s\n" ${fbcOptIn[@]} |sort |uniq))
-        mustPublishIndexImage=$fbcOptIn
-        mustSignIndexImage=$fbcOptIn
-
         if [ "$(params.hotfix)" == "true" ]; then
             echo "Hotfix build"
             fbcOptIn="false"
@@ -137,18 +145,50 @@ spec:
             fbcOptIn="false"
             mustSignIndexImage="false"
             mustPublishIndexImage="false"
+        else
+            # check if this fbc fragment is opt-in
+            echo "Fetching the image bundle from $(params.fbcFragment)..."
+            PULL_SPEC_LIST=$(opm render "$(params.fbcFragment)" | jq -r \
+            'select(.schema == "olm.bundle") | "\(.image)" | split("@")[0]' |uniq)
+            for PULL_SPEC in ${PULL_SPEC_LIST}; do
+                # make sure they query is done using the internal name instead of the public
+                PULL_SPEC="${PULL_SPEC//registry.redhat.io/registry.access.redhat.com}"
 
-        # in case of more than one image all should have the same fbc_opt_in value
-        elif [ $(wc -w <<< ${fbcOptIn[@]}) -ne 1 ]; then
-            fbcOptIn="false"
-            mustSignIndexImage="false"
-            mustPublishIndexImage="false"
+                echo "Attempting to fetch from ${FETCH_URL} to check if fragment is \`fbc_opt_in==true\`..."
+                # shellcheck disable=SC2207
+                #   (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).
+                # shellcheck disable=SC2006 # (style): Use \$(...) notation instead of legacy backticks `...`.
+                # shellcheck disable=SC2086 # (info): Double quote to prevent globbing and word splitting.
+                fbcOptIn+=(`isfbcOptIn ${PULL_SPEC}`)
+            done
+
+            # shellcheck disable=SC2207
+            #   (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).
+            # shellcheck disable=SC2068 # (error): Double quote array expansions to avoid re-splitting elements.
+            fbcOptIn=($(printf "%s\n" ${fbcOptIn[@]} |sort |uniq))
+
+            # shellcheck disable=SC2128 # (warning): Expanding an array without an index only gives the first element.
+            mustPublishIndexImage="${fbcOptIn}"
+            # shellcheck disable=SC2128 # (warning): Expanding an array without an index only gives the first element.
+            mustSignIndexImage="${fbcOptIn}"
+
+            # in case of more than one image all should have the same fbc_opt_in value
+            if [ "$(wc -w <<< "${fbcOptIn[@]}")" -ne 1 ]; then
+                fbcOptIn=("false")
+                mustSignIndexImage="false"
+                mustPublishIndexImage="false"
+            fi
         fi
 
+        # shellcheck disable=SC2128 # (warning): Expanding an array without an index only gives the first element.
         echo "Fragment has \`fbc_opt_in==${fbcOptIn}\`"
         echo "             \`mustPublishIndexImage==${mustPublishIndexImage}\`"
         echo "             \`mustSignIndexImage==${mustSignIndexImage}\`"
 
+        # these results will be used by add-fbc-contribution to control
+        # signing and publishing of the built fragment
+        # shellcheck disable=SC2128 # (warning): Expanding an array without an index only gives the first element.
+        # shellcheck disable=SC2046 # (warning): Quote this to prevent word splitting.
         jq -n -c \
            --arg fbc_opt_in "${fbcOptIn}" \
            --arg publish_index_image "${mustPublishIndexImage}" \
@@ -159,11 +199,24 @@ spec:
              "sign_index_image": $sign_index_image
             } | tostring' | tee $(results.genericResult.path)
 
+        # if it finds a build which is completed or in progress, it should exit this step and jump to
+        # the next step `s-wait-for-build-state` which will watch the build until it is completed.
+        build=$(check_previous_build "${KRB5_PRINCIPAL}" "$(params.fromIndex)" "$(params.fbcFragment)")
+        if [ -n "${build}" ]; then
+          echo "=== A previous build for this fragment was found ==="
+          echo "${build}" |tee "$(results.jsonBuildInfo.path)"
+          exit 0
+        fi
+
         # adds the json request parameters to a file to be used as input data
         # for curl and preventing shell expansion.
         json_input=/tmp/$$.tmp
         json_raw_input=/tmp/$$_raw.tmp
 
+        # shellcheck disable=SC2006 # (style): Use $(...) notation instead of legacy backticks `...`.
+        # shellcheck disable=SC2128 # (warning): Expanding an array without an index only gives the first element.
+        # shellcheck disable=SC2046 # (warning): Quote this to prevent word splitting.
+        # shellcheck disable=SC2005 # (style): Useless echo? Instead of 'echo $(cmd)', just use 'cmd'.
         cat > $json_raw_input <<JSON
         {
           "fbc_fragment": "$(params.fbcFragment)",
@@ -181,12 +234,15 @@ spec:
           if(.add_arches | length) == 0 then del(.add_arches) else . end |
           if(.build_tags | length) == 0 then del(.build_tags) else . end' ${json_raw_input} > ${json_input}
 
+        # shellcheck disable=SC2046 # (warning): Quote this to prevent word splitting.
         echo "Calling IIB endpoint" > $(results.buildState.path)
         # adds image to the index.
+        # shellcheck disable=SC2046 # (warning): Quote this to prevent word splitting.
         /usr/bin/curl -u : --negotiate -s -X POST -H "Content-Type: application/json" -d@${json_input} --insecure \
         "${IIB_SERVICE_URL}/builds/fbc-operations" |tee $(results.jsonBuildInfo.path)
 
         # checks if the previous call returned an error.
+        # shellcheck disable=SC2046 # (warning): Quote this to prevent word splitting.
         ! jq -e -r ".error | select( . != null )" $(results.jsonBuildInfo.path)
       volumeMounts:
         - name: service-account-secret


internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren, line 1: Can you bump the version string?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -83,49 +83,75 @@ spec:
       script: |
         #!/usr/bin/env bash
         #
+        # shellcheck disable=SC2207,SC2128,SC2178 # requires refactoring to fix these errors
```
johnbieren, line 86: what even are these? Can you add a little more detail

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -83,49 +83,75 @@ spec:
...
 
-          curl --negotiate -u: "${FETCH_URL}" -o $TMPFILE
+          curl --negotiate -u: "${FETCH_URL}" -o "${TMPFILE}"
           # prints "false" in case .fbc_opt_in entry is missing
-          jq -e -r '.fbc_opt_in //false' $TMPFILE && rm -f $TMPFILE
+          jq -e -r '.fbc_opt_in //false' "${TMPFILE}" && rm -f "${TMPFILE}"
+        }
+
+        # checks if there was any previous build for the same snapshot,
+        # although it should only consider resume it if is not a stage
```
johnbieren, line 106: `s/resume/resuming/`

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -83,49 +83,75 @@ spec:
...
 
         # performs kerberos authentication.
         base64 -d /mnt/service-account-secret/keytab > /tmp/keytab
 
         KRB5_TEMP_CONF=$(mktemp)
+        KRB5_PRINCIPAL=$(cat /mnt/service-account-secret/principal)
+
+        build=$(check_previous_prod_build "${KRB5_PRINCIPAL}" "$(params.fromIndex)" "$(params.fbcFragment)")
+        if [ -n "${build}" ]; then
+          echo "The fragment is already building in IIB. Resuming..."
```
johnbieren, line 127: What if the build was completed?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -83,49 +83,75 @@ spec:
...
+        # although it should only consider resume it if is not a stage
+        # build as stage builds have no reason for being resumed.
+        check_previous_prod_build() {
+
+          user="${1}"
+          from_index="${2}"
+          fbc_fragment="${3}"
+
+          # fetch only builds in progress or completed
+          curl -s "https://iib.engineering.redhat.com/api/v1/builds?user=${user}&from_index=${from_index}" | \
```
johnbieren, line 115: Shouldn't this use IIB_SERVICE_URL?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: It will get the information of  completed and exit the step. Later, in the next step it will check the status of this build and fill up the results.

johnbieren: I'm confused. This `check_previous_prod_build` part seems like the only new code to make the task idempotent. If there is a previous prod build, we say it is already building and we are resuming. Couldn't the previous prod build have already succeeded? So maybe just the comment is wrong? Because if we are resuming, why are we just calling echo then exiting without any other call?

theflockers: A think what is more confusing is  "what we have here to resume".  

In my understanding, the reasons a user would need to "retry" the build would be:

- the pipeline failed/timed out, but IIB build already finished successful on retry time;
- the pipeline failed/timed out, but IIB build is still in progress on retry time;

Only in case it is the 1st build, or a build that IIB returned  **failed**  it would need to create a new build (krb auth and IIB build call)

That said, IIB has only 3 expected status: "completed", "inprogress" ( or some similar string) and "failed".  So, the only condition it needs to continue in that  "step" is if the new function check_previous_prod_build` returns a empty string (as it ignores "failed") - unless we want to finish the pipeline and show the previous failure to the user; but I think this would for the user having a new snapshot, and FBC users use the same snapshot for different `from_index` parameters.

For "in progress" and "complete", it `exit 0` because all it needs to do is in the next "step" - the second container it spawns -, where it can run the `watch_build_state()` using the json saved in the previous step.

For the moment, the only possibility of "resume" is for prod builds, because if the RPA has the `stagedIndex` set to `true`, it has nothing to resume, because IIB does not  replace the `fromIndex` image,  so there is literally nothing to do - it won't sign and won't publish, so the only thing it can do is _run a new build_ .

Maybe I am skipping something, but anyhow the code is not properly tested. I will be doing this on Monday.

theflockers: One note:  there is a mistake indeed in the jq, which is missing specifying what it should check (non failed builds) - I missed it after trying the code, to be fixed.

For a successful IIB build, it would still need to fill up all results so the public `add-fbc-fragment` task can get the results to then publish the image - but indeed, without calling IIB for a new build.
 
What it actually does (although missing the fix on the `jq` that I  mentioned above) is in case the build is in progress or succeeded/completed, jump to the script in line 230 - which only fill up the results, without calling a new build.

The `exit 0` will exit the script , so I won`t call IIB for that purpose.

Anyhow, I am starting testing this - so if I am missing something (as it is likely) I should catch on testing.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -83,49 +83,75 @@ spec:
       script: |
         #!/usr/bin/env bash
         #
+        # shellcheck disable=SC2207,SC2128,SC2178 # requires refactoring to fix these errors
+        # shellcheck disable=SC2005 # false positive as it thinks tekton parameters might be commands
```
mmalina, line 87: What are the "false positives"? It would be better to only disable the check selectively for those lines if needed. But I wonder if it's these:

```
        "build_tags": $(echo "$(params.buildTags)"),
          "add_arches": $(echo "$(params.addArches)"),
```

It does look like the `echo` is redundant, doesn't it?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -83,49 +83,75 @@ spec:
...
-          curl --negotiate -u: "${FETCH_URL}" -o $TMPFILE
+          curl --negotiate -u: "${FETCH_URL}" -o "${TMPFILE}"
           # prints "false" in case .fbc_opt_in entry is missing
-          jq -e -r '.fbc_opt_in //false' $TMPFILE && rm -f $TMPFILE
+          jq -e -r '.fbc_opt_in //false' "${TMPFILE}" && rm -f "${TMPFILE}"
+        }
+
+        # checks if there was any previous build for the same snapshot,
+        # although it should only consider resume it if is not a stage
+        # build as stage builds have no reason for being resumed.
```
mmalina, line 107: Why can't stage builds be resumed? I'm curious.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: > What if the build succeeded last time?

@johnbieren in that case `exit 0` is called two lines below this thread, so you will never get to the "start new IIB build" curl call.

So I think it's correct, but it would be useful if we printed some more details about what we found - I think it would be good to clearly see in the logs if it was already succeeded, or even failed.

One thing that is still not clear to me is why for stage builds we always need a new build. Why can't we reuse a previous build? You say "there is nothing to do" and yet we need to run a new IIB build. Why?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: Basically because they execute a single step, which is the IIB build. They are not published or signed, although after the conversation in the biggest thread, it might be the case that the staged build  timed out/failed but IIB is still running. The user might want to know what happened, even though it won't do anything else after finish building.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: @mmalina two lines below is still in this if statement. The first line of the if says "fragment is already building, resuming." So, the comment is wrong, or the case where a build for the image is already completed is not addressed. Either way, something has to change

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: done.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: I removed the exclusion. Crossed fingers the code is still working (so far in the tests the seem to... but, lets hope to see in stage).

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: changed to use the variable

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: the code was changed to allow "resuming" staged indexes as well. PTAL.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -87,46 +87,45 @@ spec:
...
-          curl --negotiate -u: "${FETCH_URL}" -o $TMPFILE
+          curl --negotiate -u: "${FETCH_URL}" -o "${TMPFILE}"
           # prints "false" in case .fbc_opt_in entry is missing
-          jq -e -r '.fbc_opt_in //false' $TMPFILE && rm -f $TMPFILE
+          jq -e -r '.fbc_opt_in //false' "${TMPFILE}" && rm -f "${TMPFILE}"
+        }
+
+        # checks if there is any previous build for the same snapshot.
+        # in case multiple builds are found, returns only the last one.
+        check_previous_prod_build() {
```
mmalina, line 105: It's no longer only prod builds, right? So should be renamed.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: ok, thanks

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -137,57 +136,89 @@ spec:
...
+            # in case of more than one image all should have the same fbc_opt_in value
+            if [ "$(wc -w <<< "${fbcOptIn[@]}")" -ne 1 ]; then
+                fbcOptIn=("false")
+                mustSignIndexImage="false"
+                mustPublishIndexImage="false"
+            fi
         fi
 
-        echo "Fragment has \`fbc_opt_in==${fbcOptIn}\`"
+        echo "Fragment has \`fbc_opt_in==${fbcOptIn[0]}\`"
```
mmalina, line 165: This is not right. `fbcOptIn` is simply set to `"false"` for hotfix and for stage builds. But then an array for the last branch of the `if`. I would suggest to just use a string.

In the `else` branch, where you check each image separately, what you're checking is that all of them need to have `fbcOptIn=true`, right? If at least one of them has false, then you cannot publish it. So before the loop, set `fbcOptIn=true`. And in each iteration, run `isfbcOptIn` and if it's false, stop (`break`) right there - you know for sure that you cannot publish it.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: @johnbieren I think the comments make it clearer that it's in progress or completed.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -87,46 +87,45 @@ spec:
...
+        }
+
+        # checks if there is any previous build for the same snapshot.
+        # in case multiple builds are found, returns only the last one.
+        check_previous_prod_build() {
+          user="${1}"
+          from_index="${2}"
+          fbc_fragment="${3}"
+
+          # fetch only builds in progress or completed
```
mmalina, line 110: This means even failed ones, right? We would still pass them to the next step and they will the report that it failed, correct? Or if failed ones are not included, can you explain what makes them excluded? And yeah, we probably do want a new build if the previous failed, by default.

Related question: If there are two previous builds: First failed, second successful, will this correctly select the last (successful) one? I can't see how.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -87,46 +87,45 @@ spec:
...
 
-          curl --negotiate -u: "${FETCH_URL}" -o $TMPFILE
+          curl --negotiate -u: "${FETCH_URL}" -o "${TMPFILE}"
           # prints "false" in case .fbc_opt_in entry is missing
-          jq -e -r '.fbc_opt_in //false' $TMPFILE && rm -f $TMPFILE
+          jq -e -r '.fbc_opt_in //false' "${TMPFILE}" && rm -f "${TMPFILE}"
+        }
+
+        # checks if there is any previous build for the same snapshot.
+        # in case multiple builds are found, returns only the last one.
```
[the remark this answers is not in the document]
theflockers: When found, it is always a list with at least 1 element. Normally, just one, but as I was rebuilding over and over the same fbc_fragment, it ended up with more than one, so I added the index to make sure it should always get something . My initial implementation I was fetching them from `items[]`.

If the code ignore the failed, the tendency is that it will never return more than one, because we would return only in progress or complete, so the fbc_fragment will have to change to allow a new build.

You can see it here: [here](https://iib.stage.engineering.redhat.com/api/v1/builds?user=rhtap-release-iib-stage@IPA.REDHAT.COM&from_index=registry-proxy.engineering.redhat.com/rh-osbs/iib-preview-rhtap:v4.12)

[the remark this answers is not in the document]
theflockers: But maybe you are right. I might be considering the wrong one depending on what the request got. I will crosscheck this.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: > This means even failed ones, right? We would still pass them to the next step and they will the report that it failed, correct? Or if failed ones are not included, can you explain what makes them excluded? And yeah, we probably do want a new build if the previous failed, by default.

What came to my mind to re-add the "failed" ones was the following situation:
- task calls the IIB build;
- the *watch* task times out or fails for anything but IIB, so the build process is still running, but the task already finished
- after some minutes IIB build fails
- a new FBC pipeline starts, and  detects that the last build failed the user can look the failure.

The problem with this logic is that the next one would say that an existent build for the same fragment, so the user would need a new snapshot or a flag to rebuild the same snapshot.

> Related question: If there are two previous builds: First failed, second successful, will this correctly select the last (successful) one? I can't see how.

IIB seems to sort the builds, so the latest is 0. One option would sort the result before select the index. I will ask the IIB fellows if we can trust the output as "sorted".

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: this is one of the reasons I wanted initially to add the shellcheck exclusions. I would prefer rewrite the full fbc_opt_in part because this is looking ugly. I really can't remember why eventually it would return more than 1 items in the array, and I was a bit afraid to break the builds by trying to fix this now.

Anyhow the change done is basically trusting shellcheck that informed that the above code would return the first element of the array in all cases, so that why I pinned the 0, so it passes (in theory) without change the behavior.

What would you suggest? Should it fix it properly in this ticket or comment out this and create a fix ticket for it? Both works for me.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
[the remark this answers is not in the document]
theflockers: Another comment. The link above uses to search the "from_index"  used by e2e, so the fbc_fragment is always the same. Not sure how frequent this would happen for actual users.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: AFAIK, if a build was successful and the `user`, `from_index` and `fbc_fragment` is the same (I guess these are the unique params from your comment), we should not rebuild the fragment. That is the point of idempotent

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -87,46 +87,45 @@ spec:
...
           [ -z "${TAG}" ] && FETCH_URL=${FETCH_URL%/tag*}
 
-          curl --negotiate -u: "${FETCH_URL}" -o $TMPFILE
+          curl --negotiate -u: "${FETCH_URL}" -o "${TMPFILE}"
           # prints "false" in case .fbc_opt_in entry is missing
-          jq -e -r '.fbc_opt_in //false' $TMPFILE && rm -f $TMPFILE
+          jq -e -r '.fbc_opt_in //false' "${TMPFILE}" && rm -f "${TMPFILE}"
+        }
+
+        # checks if there is any previous build for the same snapshot.
```
johnbieren, line 103: This only handles single component applications, right? Maybe it makes sense to `s/snapshot/fragment/`.. up to you

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: Does this relate to making the task idempotent? I am getting confused

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -137,57 +136,89 @@ spec:
...
-          "add_arches": `echo $(params.addArches)`,
-          "overwrite_from_index": ${fbcOptIn},
+          "build_tags": $(params.buildTags),
+          "add_arches": $(params.addArches),
+          "overwrite_from_index": ${fbcOptIn[0]},
           "overwrite_from_index_token": "${IIB_OVERWRITE_FROM_INDEX_USERNAME}:${IIB_OVERWRITE_FROM_INDEX_TOKEN}"
         }
         JSON
 
+        cat "${json_raw_input}"
```
johnbieren, line 206: testing leftover?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -137,57 +136,89 @@ spec:
...
         # filtering out empty params
         jq -r '
           if .overwrite_from_index == false then del(( .overwrite_from_index, .overwrite_from_index_token)) else . end |
           if(.add_arches | length) == 0 then del(.add_arches) else . end |
           if(.build_tags | length) == 0 then del(.build_tags) else . end' ${json_raw_input} > ${json_input}
 
-        echo "Calling IIB endpoint" > $(results.buildState.path)
+        cat "${json_input}"
```
johnbieren, line 213: testing leftover?

johnbieren (commented):
Making the task idempotent is a pretty big change. It might be nice to narrow this PR to doing just that and leave shellcheck/other improvements to another PR

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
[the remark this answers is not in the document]
mmalina: Thanks for the link to iib. Ok, so at least I can see that the order is latest to oldest. So yeah, we want something from the top. But I still don't see how this filter could work properly:

```
jq -r --arg fbc_fragment "${fbc_fragment}" '.items[0] |select(.fbc_fragment==$fbc_fragment)'
```

You only select the 1 latest build from the results. And then you do the select - so if it's the one with your fragment, the build json is the result, if not, the result is empty. And it doesn't even do anything to skip failed builds - how is that supposed to work?

You will need something like:
```
jq '[.items[] |select(.fbc_fragment=="quay.io/hacbs-release-tests/fbc-sample-repo@sha256:857814679c1deec5bc5d6d8064832b4c0af40dcb07dad57c48f23e5ba6926aed" and .state != "failed")][0]'
```

(Not sure what the state value for failed is.)

And then you need to check if that result is empty or not - if not, you can use that build, otherwise you need a new one.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: Yeah, see the other thread - they are sorted from newest to latest it seems. And you should be able to filter out the failed ones from the result.

I understand the use case for showing the failed one, but I don't think we need to worry about that - I think it should be pretty rare that it would fail, but take so long for the failure that we would time out. And even if that happens, that might mean that IIB is overloaded and so you can't really move forward with your release until it can finish builds in a reasonable time frame.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: OK, if you only touched this part to make shell check happy, I'm fine adding shellcheck ignores (but only for the affected lines, not at the top) and creating a follow up jira to fix it later.

But it definitely is not right to treat it as an array here when we can clearly see that in some cases it will just be a string.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: removed

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: removed.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: changed to `fbc_fragment`.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: changed back to consider only *in_progress* and *completed* builds. Failed ones will build again by default.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: renamed.

theflockers:
> Making the task idempotent is a pretty big change. It might be nice to narrow this PR to doing just that and leave shellcheck/other improvements to another PR

NP. I will change back the shellcheck parts and add them later in separate.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: Do we need a follow up ticket? Is the existing one about fixing shellcheck issues in this task not enough?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -157,19 +179,29 @@ spec:
              "fbc_opt_in": $fbc_opt_in,
              "publish_index_image": $publish_index_image,
              "sign_index_image": $sign_index_image
-            } | tostring' | tee $(results.genericResult.path)
+            } | tostring' | tee "$(results.genericResult.path)"
+
+        # if it finds a build which is completed or in progress, it should exit this step and jump to
+        # the next step `s-wait-for-build-state` which will watch the build until it is completed.
+        build=$(check_previous_build "${KRB5_PRINCIPAL}" "$(params.fromIndex)" "$(params.fbcFragment)")
+        if [ -n "${build}" ]; then
```
johnbieren, line 187: You have `// empty` on the `jq` command in `check_previous_build`. Won't this always be true?
```
$ foo=$(jq '.foo // empty' <<< "{}")
$ if [ -n $foo ] ; then echo hi ; fi
hi
```

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -87,46 +87,45 @@ spec:
...
         echo "${KRB5_CONF_CONTENT}" > "${KRB5_TEMP_CONF}"
         export KRB5_CONFIG="${KRB5_TEMP_CONF}"
+
         export KRB5_TRACE=/dev/stderr
 
-        /usr/bin/kinit -V $(cat /mnt/service-account-secret/principal) -k -t /tmp/keytab
+        /usr/bin/kinit -V "${KRB5_PRINCIPAL}" -k -t /tmp/keytab
 
         set -x
-        # check if this fbc fragment is opt-in
```
johnbieren, line 114: It isn't clear to me how this block of code being moved is related to making the task idempotent

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: can be as well. I will add some more info to the current ticket then.  Because the logic needs  a bit of change, not only the syntax  check.

internal-services/catalog/~shellcheck.diff
johnbieren, line 1: Accidental add?

theflockers: ah sure. Thanks for noticing.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: gotcha, thanks

johnbieren (approved):
Looks okay to me with my limited knowledge of the specifics of this task :slightly_smiling_face:

johnbieren:
The checkton check should be overridden until the checkton ticket for this is fixed. The other linter errors should be addressed though

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: cool. Just tested it and it is working fine. I am resolving it, thanks!

mmalina:
> FYI @mmalina I removed all stuff regarding the shellcheck. It will be handled up next in https://issues.redhat.com/browse/RELEASE-1099

OK, so can you add a comment there? The Jira has some history and last comment is from November 22 a points to this PR, so it's confusing.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: sounds good

mmalina:
> The checkton check should be overridden until the checkton ticket for this is fixed. 

OK. Currently yamllint and gitlint fail as well, so those need to be fixed.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -137,18 +136,37 @@ spec:
...
-        # in case of more than one image all should have the same fbc_opt_in value
-        elif [ $(wc -w <<< ${fbcOptIn[@]}) -ne 1 ]; then
-            fbcOptIn="false"
-            mustSignIndexImage="false"
-            mustPublishIndexImage="false"
+                echo "Attempting to fetch from ${FETCH_URL} to check if fragment is \`fbc_opt_in==true\`..."
+                IFS=" " read -r -a fbcOptIn <<< "$(isfbcOptIn "${PULL_SPEC}")"
+            done
+
+            IFS=" " read -r -a fbcOptIn <<< "$(printf "%s\n" "${fbcOptIn[@]}" |sort |uniq)"
```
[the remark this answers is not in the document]
theflockers: Please take a look in the new [PR](https://github.com/hacbs-release/app-interface-deployments/pull/236) (although it is a draft as it will need rebase) but I made a few changes in this part there, and I also added some comments to explain the logic. 
As Johnny mentioned, I won´t change the logic of the fbcOptIn in this PR. The small code moved around was just to make sure the `fbcOptIn` function would run only when necessary.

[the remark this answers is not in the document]
theflockers: Anyhow I will take in consideration the comment above in the new PR.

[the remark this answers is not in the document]
mmalina: OK, but why was this:
```
fbcOptIn=($(printf "%s\n" ${fbcOptIn[@]} |sort |uniq))
```
changed to this?
```
IFS=" " read -r -a fbcOptIn <<< "$(printf "%s\n" "${fbcOptIn[@]}" |sort |uniq)"
```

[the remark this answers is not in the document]
mmalina: > Please take a look in the new https://github.com/hacbs-release/app-interface-deployments/pull/236

But it still has this weird array saving, instead of just one value. I will comment there.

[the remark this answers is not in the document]
theflockers: Hum.. this should not be here  anymore. I missed removing it when reverting to the original code. I will follow up there once I revert this piece of code.

theflockers:
> > The checkton check should be overridden until the checkton ticket for this is fixed.
> 
> OK. Currently yamllint and gitlint fail as well, so those need to be fixed.

All good now. PTAL @mmalina

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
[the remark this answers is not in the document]
mmalina: ok, let me know what it's reverted to the original

mmalina (approved):

johnbieren (approved):

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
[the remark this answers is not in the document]
theflockers: resolving this one as the PR is approved.

PR #233, opened 2024-11-26
chore(RELEASE-1040): remove file-updates

scoheb:
- migration performed in RELEASE-1283


scoheb:
force merging since the RHOAI needs to start testing their add-on in stage environment

PR #234, opened 2024-11-26
Revert "chore(RELEASE-1040): remove file-updates"

scoheb:
Reverts hacbs-release/app-interface-deployments#233


scoheb:
merged too fast. need these for e2e tests

PR #236, opened 2024-11-27
fix(RELEASE-1099): lint issue on iib-add-fbc-fragment-to-index-image

theflockers:
this PR fixes the lint issues in the taskiib-add-fbc-fragment-to-index-image-task. In
addition, adjusts the fbc_opt_in code tobe more clear.

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.3.2"
+    app.kubernetes.io/version: "0.3.3"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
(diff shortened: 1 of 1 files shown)


theflockers:
follow up PR of https://github.com/hacbs-release/app-interface-deployments/pull/231

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -108,69 +108,79 @@ spec:
...
+        PULL_SPEC_LIST=$(opm render "$(params.fbcFragment)" | jq -r \
         'select(.schema == "olm.bundle") | "\(.image)" | split("@")[0]' |uniq)
         for PULL_SPEC in ${PULL_SPEC_LIST}; do
             # make sure they query is done using the internal name instead of the public
-            PULL_SPEC=$(sed  's/registry.redhat.io/registry.access.redhat.com/g' <<< $PULL_SPEC)
+            PULL_SPEC="${PULL_SPEC//registry.redhat.io/registry.access.redhat.com}"
 
             echo "Attempting to fetch from ${FETCH_URL} to check if fragment is \`fbc_opt_in==true\`..."
-            fbcOptIn+=(`isfbcOptIn ${PULL_SPEC}`)
+            IFS=" " read -r -a fbcOptIn <<< "$(isFBCOptIn "${PULL_SPEC}")"
```
mmalina, line 124: Why do we need to save an array? Isn't it enough to know that at least one pull spec was false?

Originally commented here: https://github.com/hacbs-release/app-interface-deployments/pull/231#discussion_r1865754661

You could just do something like this:
```
fbcOptIn=true
for PULL_SPEC in ${PULL_SPEC_LIST}; do
    # make sure they query is done using the internal name instead of the public
    PULL_SPEC="${PULL_SPEC//registry.redhat.io/registry.access.redhat.com}"
    echo "Attempting to fetch from ${FETCH_URL} to check if fragment is \`fbc_opt_in==true\`..."
    if [ "$(isfbcOptIn "${PULL_SPEC}")" = false ]; then
        fbcOptIn=false
        
        # This is optional - it should be enough to know that one is false, but maybe we want check all of them 
        # to print to the log - but currently it doesn't print it anyway I think.
        break 
    fi
done
mustPublishIndexImage="${fbcOptIn}"
mustSignIndexImage="${fbcOptIn}"
```

theflockers: nice. I will try this change. thanks.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -137,67 +130,50 @@ spec:
...
+                fi
             done
-
-            # shellcheck disable=SC2207
-            #   (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).
-            # shellcheck disable=SC2068 # (error): Double quote array expansions to avoid re-splitting elements.
-            fbcOptIn=($(printf "%s\n" ${fbcOptIn[@]} |sort |uniq))
-
-            # shellcheck disable=SC2128 # (warning): Expanding an array without an index only gives the first element.
+            mustOverwriteFromIndexImage="${fbcOptIn}"
```
mmalina, line 157: Ok, so you basically renamed `fbcOptIn` to `mustOverwriteFromIndexImage`. Wouldn't it make sense to use it even here from the start? 

```
...
else
  mustOverwriteFromIndexImage=true
  for ...
    if [ "$(isFBCOptIn "${PULL_SPEC}")" = "false" ]; then
      mustOverwriteFromIndexImage=false
```

mmalina: I think this will actually make it clearer - there are potentially multiple images checked, so with the different name, we are saying: if isFBCOptIn is false for an image, we need to override the index image.

theflockers: I changed this again so we can keep the fbc_opt_in as it was found in the image as  per our conversation  [here](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1733483922183879?thread_ts=1733474423.638719&cid=C04PZ7H0VA8). 
 
- The IsFBCOptIn check is back top, so hotfix and stage would also check the flag.
- the fbcOptIn value is not replaced anymore.
 
 Can you PTAL?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren, line 1: Can you bump the z of the version string?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
mmalina: Ah, I see, I didn't realize there were two different things - one is the result and one is overwrite_from_index parameter for the iib call. Yep, it makes sense now.

mmalina (approved):
LGTM once Johnny's comment is addressed.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: done.

johnbieren (approved):

johnbieren:
Needs commits squashed

mmalina (approved):

PR #239, opened 2024-12-03
feat(RELEASE-1245): Make publish-index-image task idempotent

happybhati:


internal-services/catalog/publish-index-image-task.yaml
```diff
@@ -49,14 +49,53 @@ spec:
...
+        AUTH_PARAM=()
+        # Do not authenticate if the source is Red Hat's "registry-proxy" which is unauthenticated.
         if [[ ! "$(params.sourceIndex)" =~ ^registry-proxy(\-stage)?.engineering.redhat.com ]]; then
             AUTH_PARAM=("--src-creds" "${SOURCE_INDEX_CREDENTIAL}")
         fi
+        TARGET_AUTH="--dest-creds ${TARGET_INDEX_CREDENTIAL}"
+
+        # Check if the target image exists
+        echo "Checking if target image exists: $(params.targetIndex)"
+        set +e
```
theflockers, line 66: You can do this without changing the flag, if the command is inside an `if` or using `||` in the end of the command, for instance: 

```
if ! my command; then
  ERROR=1
fi
```
or
```
my command ||  ERROR=$?
```

internal-services/catalog/publish-index-image-task.yaml
```diff
@@ -49,14 +49,53 @@ spec:
...
+            "docker://$(params.sourceIndex)" \
+            --format '{{.Digest}}')
+
+        if [ $INSPECT_EXIT_CODE -eq 0 ]; then
+            if [ "$SOURCE_DIGEST" == "$TARGET_DIGEST" ]; then
+                echo "Image already exists in target registry with the same digest." \
+                     "Skipping copy." | tee "$(results.requestMessage.path)"
+                exit 0
+            else
+                echo "Image exists in target registry but digests do not match." \
```
theflockers, line 89: The targetIndex is the catalog and will always exist - except for hotfix, which is always different.
So only checking if the digest is the same is enough.

internal-services/catalog/publish-index-image-task.yaml
happybhati: Thanks for suggestion, fixed!

internal-services/catalog/publish-index-image-task.yaml
happybhati: Ah got it, fixed now, as you suggested. PTAL!

internal-services/catalog/publish-index-image-task.yaml
```diff
@@ -49,14 +49,46 @@ spec:
...
+        SOURCE_DIGEST=$(skopeo inspect \
+            --tls-verify=false \
+            "${AUTH_PARAM[@]}" \
+            "docker://$(params.sourceIndex)" \
+            --format '{{.Digest}}')
+
+        echo "Getting target image digest: $(params.targetIndex)"
+        if TARGET_DIGEST=$(skopeo inspect \
+            --tls-verify=false \
+            "${TARGET_AUTH}" \
```
mmalina, line 74: I don't think this will work. This would result in one cli argument `"--dest-creds ${TARGET_INDEX_CREDENTIAL}"` - you want two. So you should use an array - same way `AUTH_PARAM` is done. And since you already create this, I would suggest to also use the array in the `skopeo copy` command at the end.

mmalina: Also, I would suggest better names for the variables: `SOURCE_AUTH_ARGS` and `TARGET_AUTH_ARGS`.

internal-services/catalog/publish-index-image-task.yaml
```diff
@@ -49,14 +49,46 @@ spec:
...
+            "docker://$(params.targetIndex)" \
+            --format '{{.Digest}}' 2>/dev/null); then
```
mmalina, line 76: Why `2>/dev/null`? I don't think it's a good idea. It's much better to see the error in the log if something happens. It won't have any effect on the variable, because that will only store stdout, not stderr.

internal-services/catalog/publish-index-image-task.yaml
```diff
@@ -49,14 +49,46 @@ spec:
...
+            echo "Target image exists."
+            if [ "$SOURCE_DIGEST" == "$TARGET_DIGEST" ]; then
+                echo "Image already exists in target registry with the same digest." \
+                    "Skipping copy." | tee "$(results.requestMessage.path)"
+                exit 0
+            else
+                echo "Image exists in target registry but digests do not match." \
+                    "Proceeding to copy the image." | tee "$(results.requestMessage.path)"
```
mmalina, line 84: The results file will get overridden on line 69/101, so there's no use writing anything to it here. If you wanted to have both, you could use `tee -a` at the end.

internal-services/catalog/publish-index-image-task.yaml
```diff
@@ -49,14 +49,46 @@ spec:
...
+            --format '{{.Digest}}')
+
+        echo "Getting target image digest: $(params.targetIndex)"
+        if TARGET_DIGEST=$(skopeo inspect \
+            --tls-verify=false \
+            "${TARGET_AUTH}" \
+            "docker://$(params.targetIndex)" \
+            --format '{{.Digest}}' 2>/dev/null); then
+            echo "Target image exists."
+            if [ "$SOURCE_DIGEST" == "$TARGET_DIGEST" ]; then
```
mmalina, line 78: @theflockers I wonder how likely this is. Let's say you run your release pipeline the first time and then the second time. The digest will only be the same if nobody else published anything between the two runs, right? Shouldn't we rather inspect the published index image and see if our fragment is already in the fbc catalog inside the image? It would definitely be harder to do, but it seems to me that it's the only way to really know if we need publish it, no?

theflockers: As of today it is quite likely to happen because we don't have many prod fbc releases yet, but once more and more operators are built in konflux, indeed, it will be less likely to happen.

So it might make sense checking if the component already exists in the catalog,  but it would add some overhead in the taskrun time  - it would jump from seconds to (today) up to 5 minutes - but we we have really long runs, this overhead can be neglected.

But having in mind the scenario you mentioned, I thought about another situation that could happen:

-  if the image to be pushed has different digest but it is older than the current published catalog  (if someone resumes a timed out run, it might be the case that someone else had a successful run that published the catalog, and at the time of the publishing my catalog is older). 

In this case, it should skip publishing and tell the user to generate a new snapshot (in comprehensive words).

mmalina: > In this case, it should skip publishing and tell the user to generate a new snapshot (in comprehensive words).

Interesting problem. I might be missing something. But is it really necessary to have a new snapshot? Wouldn't the user normally expect that a release retry should fix the issue, e.g. the index image will be generated again and then published? But of course then we have a problem how to handle it on our side :)

internal-services/catalog/publish-index-image-task.yaml
```diff
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: task-publish-index-image
   labels:
-    app.kubernetes.io/version: "0.1.1"
+    app.kubernetes.io/version: "0.1.2"
```
johnbieren, line 7: Making something idempotent is a big enough change to go to 0.2.0 I'd say

internal-services/catalog/publish-index-image-task.yaml
johnbieren: I am confused. This task runs a `skopeo copy`. The change here checks if the target digest equals source digest. If so, don't push. That sounds simple enough to me. What are we saying is wrong?

internal-services/catalog/publish-index-image-task.yaml
johnbieren: Yeah I don't think this line should be put in the results. The one from line 69/101 is better

internal-services/catalog/publish-index-image-task.yaml
mmalina: I don't think it's as simple as that - my understanding is that what matters is if your fragment (if that's the right word here) is already included in the fbc catalog in the published index image. The actual digest of that image is not very significant - other than if it's the same, then of course you can say with certainty that your fragment is already there - it's the same image. Leandro, correct me please if I'm wrong here :)

happybhati: Thank you all for the detailed discussion. I understand that relying solely on the digest comparison is a quick and lightweight check, but as mentioned, it doesn't guarantee that our specific fragment is included if the digest differs. Implementing a deeper, content-based verification would be more accurate but also adds significant overhead and complexity.

Given the current volume of releases, the digest-based approach might still be acceptable as a first step. Over time, as the release frequency grows and scenarios become more complex, we can revisit this logic to implement a more content-aware check. This could mean extracting and examining the FBC catalog inside the target image to confirm whether our fragment is already present.

I’m open to either approach. If the team feels it's worth the extra runtime to ensure absolute correctness now, I can look into the more detailed verification method (of course, with Leandro's help and expertise in FBC :). Otherwise, we can proceed with the digest-based shortcut and schedule a future enhancement once the frequency of updates makes it necessary.

What do you think about starting simple and evolving the approach as needed?

theflockers: Right. The correct then is doing this even before calling IIB, so the publishing does not need to take this in consideration. I will move the IIB ticket back to in progress and fix this.

theflockers: Sorry my last answer was based in the discussion from earlier yesterday (the new interactions were not there when I answered).

Both cases (comparing digests or inspecting if the new fragment already exists) in theory should be enough to tell if the image should be published or not. The problem for me lies in other place:

IIB has a mechanism to control concurrent builds, so if you have 2 parallel builds, both fragments will be present in the catalog. 

So lets say, you have 2 simultaneous release: 1 timed out, 1 finished successfully, hence only the  one thoese published the fragment.
Then you have a 3rd (unrelated)  fbc release. It succeeded, so fragment is published.

So the 1st user notices that his build has timed out and retriggers the release. As the fragment is already built, it only continues from where it stopped. The IIB internal pipeline gives the old build catalog to be published.
That catalog will differ and remove the fragment of the 3rd release.

So I think the solution here is: Even if the fragment was already built, it should check if the catalog that resulted from that release is not older than the current one. If yes, a new IIB build should be called.

This way (i think) this task will only need to check if the digests differs, because it can only publish new catalogs.

theflockers: I created it as a [bug](https://issues.redhat.com/browse/RELEASE-1338) ticket to handle this because the PR for the iib-add-fbc-fragment task is already merged.

johnbieren: I'm not sure we should aim to solve 2 simultaneous releases at the same time here. That is already an issue all over the place. See https://issues.redhat.com/browse/RELEASE-252
The idempotent goal AIUI is if you have a release A for snapshot X and release A fails for any reason so you create release B for snapshot X, the task checks if it already ran for this snapshot and does not do it again if it did. So, I am not following why checking the digest is not enough. How does the digest change for the same snapshot? I also don't have much knowledge about fbc fragments so I am very likely missing something

mmalina: > I'm not sure we should aim to solve 2 simultaneous releases here. That is already an issue all over the place. See https://issues.redhat.com/browse/RELEASE-252

I hope you don't mean we shouldn't allow two simultaneous release pipelines across all of Konflux. But IIUC, that's what Leandro talked about - not limited to one app.

> So, I am not following why checking the digest is not enough. How does the digest change for the same snapshot?

We're talking about the digest of the index image. There is one published index image for all operators. If there was one index image per fbc app/component, then sure, but that's not the case.

But yeah, I don't have the full knowledge either, so if Leandro things the digest check should be enough (i.e. it's not likely we could still skip the publish if the digest differs) then let's stick to just that.

theflockers: This [PR](https://github.com/hacbs-release/app-interface-deployments/pull/253) should avoid that a wrong catalog would get here. If it was allowed to be resumed in the previous step, then it should only publish if the digest differs.

johnbieren: I was talking about the same app

johnbieren:
I know you are out for a few weeks, but once you are back, you will have to open a new PR for this as it is being migrated here https://github.com/konflux-ci/release-service-catalog/pull/736
Sorry about that

PR #241, opened 2024-12-03
refactor(RELEASE-1191): add pyxis/umb specific keys to configmaps

johnbieren:
This commit modifies the signing configMaps to have pyxis and umb specific ssl keys. They are just copied for now so nothing breaks. The existing ones will be removed at a later date so that only the PYXIS_ and UMB_ ssl keys remain.


mmalina (approved):
This looks good. But keep in mind that you'll also need to make this change for the configMaps that are deployed in the rhtap-releng workspace in Konflux clusters for the internal-pipeline use case - not sure where those are defined.

But this PR is only affecting stage deployment. And internal-pipeline is only working in production Konflux, so maybe that can wait for when you create this PR for the stable branch here.

scoheb (approved):

PR #242, opened 2024-12-04
chore: re-enabling debug

theflockers:
this PR re-enables the debug in InternalServicesConfig. A following up PR will add a cronjob to keep only the last day of internalrequests.


happybhati (approved):

johnbieren (approved):

johnbieren:
I think we should do the same thing for stage (debug true + cronjob)

theflockers:
> I think we should do the same thing for stage (debug true + cronjob)

The debug is already true in stage. Only the cronjob is missing, so I  will create a PR for stage cronjob as well.

PR #243, opened 2024-12-04
chore: add cronjob to clean up pipelineruns

theflockers:
this PR adds a new cronjob to the stonesoup-int-srvc namespace to clean up the last day pipelineruns left of finished internal-requests.

### DIFF
--- a/internal-services/config/cronjob/cleanup-internal-requests-pipelineruns.yaml
+++ b/internal-services/config/cronjob/cleanup-internal-requests-pipelineruns.yaml
@@ -1,0 +1,89 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cleanup-internal-requests-pipelineruns
+  namespace: stonesoup-int-srvc
+spec:
+  schedule: "10 03 * * *"
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          serviceAccountName: controller-manager
+          volumes:
+            - name: temp-directory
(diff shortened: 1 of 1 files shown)


johnbieren (commented):
Are we sure we are the only plrs in this namespace? We don't need to use a label selector for release or anything?

theflockers:
> Are we sure we are the only plrs in this namespace? We don't need to use a label selector for release or anything?

As off today yes. We are the only ones using this. But sure, I can make it more specific just in case.

theflockers:
btw, just noticed I did not put it in the right directory.

theflockers:
This needs to be added as well in app-interface, so I am adding both envs as the other configs found in the `config/` directory, although in app-interface they will differ `main`  and `stable` branches.

johnbieren (commented):
I know you are following other examples in this repo, but I don't really agree with them - why not to just have one (not staging vs production) and then have the sre mapping for stage clusters consume the main branch and the sre mapping for prod clusters consume the stable branch? Similar to how we do for the tekton resources in this repo (but dissimilar to how we have the signing configmaps set up)

theflockers:
> I know you are following other examples in this repo, but I don't really agree with them - why not to just have one (not staging vs production) and then have the sre mapping for stage clusters consume the main branch and the sre mapping for prod clusters consume the stable branch? Similar to how we do for the tekton resources in this repo (but dissimilar to how we have the signing configmaps set up)

The sre config in app-interface is already using different branches. I will leave a single one then - i like it better too.

johnbieren (approved):

johnbieren:
Can you squash the commits?

theflockers:
> Can you squash the commits?

done

PR #245, opened 2024-12-04
fix(KFLUXSPRT-794): pass content-gateway token as env var

swickersh:
- Change name of CGW_TOKEN to CGW_PASSWORD
  - This is actually a token, no pw, but the pubtools library expects this environment variable name
  - Relevant PR to pubtools: https://github.com/release-engineering/pubtools-content-gateway/pull/38
  - Accompanies this open PR: 
      - https://github.com/konflux-ci/release-service-utils/pull/334

### DIFF
--- a/internal-services/catalog/pulp-push-disk-images-task.yaml
+++ b/internal-services/catalog/pulp-push-disk-images-task.yaml
@@ -101,7 +101,7 @@ spec:
             secretKeyRef:
               name: $(params.cgwSecret)
               key: username
-        - name: CGW_TOKEN
+        - name: CGW_PASSWORD
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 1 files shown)


swickersh:
@scoheb Can you please double check this. I don't interact with this workflow usually. Thanks!

scoheb (approved):

scoheb:
I just realized that. I asked @swickersh in Slack.

mmalina:
This is not the first time I see something like this. I would suggest to either open the PR as Draft when it's not ready yet, or create a thread about the missing thing so that it's clear it needs to be addressed first - the same situation is in another PR, so I opened a thread there: https://github.com/konflux-ci/release-service-catalog/pull/724/files#r1871048033

PR #247, opened 2024-12-04
refactor(RELEASE-1191): add pyxis/umb specific keys to configmaps

johnbieren:
This commit modifies the signing configMaps to have pyxis and umb specific ssl keys. They are just copied for now so nothing breaks. The existing ones will be removed at a later date so that only the PYXIS_ and UMB_ ssl keys remain.


johnbieren:
Same as https://github.com/hacbs-release/app-interface-deployments/pull/241

mmalina (approved):

PR #250, opened 2024-12-06
chore: add simple-signing-pipeline

theflockers:
this PR adds the simple-sign-pipeline to the stable as the development branch is broken because this
pipeline is missing.

### DIFF
--- a/internal-services/catalog/sign-image-pipeline.yaml
+++ b/internal-services/catalog/sign-image-pipeline.yaml
@@ -1,0 +1,118 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: simple-signing-pipeline
+  labels:
+    app.kubernetes.io/version: "0.1"
+  annotations:
+    tekton.dev/pipelines.minVersion: "0.12.1"
+    tekton.dev/tags: release
+spec:
+  description: >-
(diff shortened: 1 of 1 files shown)


internal-services/catalog/sign-image-pipeline.yaml
mmalina, line 1: I don't think this is the correct version. If anything, it should be the same one that is deployed to staging: https://github.com/konflux-ci/release-service-catalog/blob/development/pipelines/simple-signing-pipeline/simple-signing-pipeline.yaml - note that it has a combined request-and-upload-signature task.

johnbieren: +1 (staging and development are the same right now... in theory we should be adding the one in `internal` but that is the same as development and staging)
Thanks for opening this!

johnbieren: Opened https://github.com/hacbs-release/app-interface-deployments/pull/251 because I am impatient and it breaks development :slightly_smiling_face:

johnbieren:
Done here https://github.com/hacbs-release/app-interface-deployments/pull/251

internal-services/catalog/sign-image-pipeline.yaml
theflockers: Np. My bad not seeing the interaction here.

PR #252, opened 2024-12-06
chore: remove simple-signing-pipeline

johnbieren:
It is in release-service-catalog production.


johnbieren:
To be merged with December 11 release-service-catalog promotion

theflockers (approved):

PR #253, opened 2024-12-09
fix(RELEASE-1338): new build for old catalog index

theflockers:
this PR make the iib-add-fbc-fragment to require a new IIB build when the resumed catalog is older than the upstream one to ensure the catalog integrity.

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.3.3"
+    app.kubernetes.io/version: "0.4.0"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
@@ -19,6 +19,10 @@ spec:
       type: string
       description: ->
         Index image (catalog of catalogs) the FBC fragment will be added to
+    - name: targetIndex
+      type: string
+      description: ->
+        Target index is the pullspec the FBC catalog will be pushed to
     - name: buildTags
       type: string
       description: ->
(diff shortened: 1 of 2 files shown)


johnbieren:
This is a new required parameter, so it is a breaking change to the task creating the InternalRequest, no?

theflockers:
> This is a new required parameter, so it is a breaking change to the task creating the InternalRequest, no?

luckily it is not a breaking change because this parameter is already sent by the managed pipeline task. I suspect (did not look tbh) this parameter was dropped some time ago but it was not changed in the managed one. 

https://github.com/konflux-ci/release-service-catalog/blob/1b1616f3bb7e14b2a85cb6d2989ab18b228b2fe9/tasks/add-fbc-contribution/add-fbc-contribution.yaml#L136C1-L137C1

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -109,9 +113,24 @@ spec:
           fbc_fragment="${3}"
 
           # fetch only builds in progress or completed
-          curl -s "${IIB_SERVICE_URL}/builds?user=${user}&from_index=${from_index}" | \
+          build=$(curl -s "${IIB_SERVICE_URL}/builds?user=${user}&from_index=${from_index}" | \
             jq --arg fbc_fragment "${fbc_fragment}" \
-              '[.items[] |select(.fbc_fragment==$fbc_fragment and .state!="failed")][0] // empty'
+              '[.items[] |select(.fbc_fragment==$fbc_fragment and .state!="failed")][0] // empty')
+
+          if [ "$(jq '.state' < "${build}")" = "complete" ]; then
```
johnbieren, line 120: Do you need `-r` on this `jq`?

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
```diff
@@ -109,9 +113,24 @@ spec:
...
+            newCatalogCreatedDate="$(date --date "$(skopeo inspect --config "${indexImageResolved}" | \
+              jq -r .created)" "+%s")"
+            upstreamCatalogCreatedDate="$(date --date "$(skopeo inspect --config "$(params.targetIndex)" | \
+              jq -r .created)" "+%s")"
+            # checks if the index_image_resolved in the previous completed build is newer
+            # than the upstream catalog index.
+            # in case the new catalog index is older than the upstream, a new build is
+            # required to assure the catalog integrity.
+            if [ "${newCatalogCreatedDate}" -gt "${upstreamCatalogCreatedDate}" ]; then
+              echo "${build}"
```
johnbieren, line 131: All we want to do here is echo? The comment makes it sound like we would exit out of the step or something

theflockers: The function was designed (in a previous PR) to output an empty string in case there is no existent build, or the json string of the build in case it can be resumed. 

Considering this, the logic here basically checks if the creationDate of the found image in the build is newer than the one in the published catalog, if so it returns the build, so it can be resumed.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
theflockers: fixed.

internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
johnbieren: Sorry, I didn't expand my UI and realize we were inside a function

johnbieren (approved):

PR #255, opened 2024-12-10
fix: update utils image in push-disk-images task

johnbieren:
The script called in the pulp-push-disk-images task started failing because `packaging` was not explicitly installed in the utils image. Now that it is, the utils image is bumped to fix the error.


seanconroy2021 (approved):
LGTM

PR #258, opened 2024-12-12
fix: update utils image in push-disk-images task

mmalina:
The image was updated recently to get a fix for a missing packaging python module:
https://github.com/hacbs-release/app-interface-deployments/pull/256

But the new image introduced a regression in the
select-oci-auth script. So now it's fixed in this new image: https://github.com/konflux-ci/release-service-utils/pull/341


johnbieren (approved):

johnbieren:
This was missed when I migrated to release-service-catalog, so the bug is back. Fixed here https://github.com/konflux-ci/release-service-catalog/pull/768

PR #270, opened 2025-01-14
chore: enable workspace for iib pipeline and task

theflockers:
this PR enables using a workspace to save the files used by the fbc (iib) task. this also allows using files in the tests that will be added to the task.

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.4.0"
+    app.kubernetes.io/version: "0.5.0"
   annotations:
     tekton.dev/pipelines.minVersion: "0.12.1"
     tekton.dev/tags: release
@@ -46,6 +46,10 @@ spec:
       type: string
       default: "false"
       description: Whether this build is for a staged index build
+    - name: subdirectory
+      description: Subdirectory inside the workspace to be used
+      type: string
+      default: ""
   results:
     - name: jsonBuildInfo
       description: JSON build information for the requested build
@@ -59,6 +63,9 @@ spec:
       description: The link to the log from the IIB request
     - name: exitCode
       description: The exit code from the task
+  workspaces:
+    - name: data
+      description: workspace to read and save files
   steps:
     - name: s-add-fbc-fragment-to-index-image
       image: >-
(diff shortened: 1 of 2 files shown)


johnbieren:
I don't think we should add a workspace to the task just to help unit tests. Adding a workspace is expensive (we already have tickets open to improve performance due to workspaces). Martin and I actually talked about it in the last week here https://github.com/konflux-ci/release-service-catalog/pull/761#discussion_r1913110859

theflockers:
> I don't think we should add a workspace to the task just to help unit tests. Adding a workspace is expensive (we already have tickets open to improve performance due to workspaces). Martin and I actually talked about it in the last week here [konflux-ci/release-service-catalog#761 (comment)](https://github.com/konflux-ci/release-service-catalog/pull/761#discussion_r1913110859)

If I understood correctly the issue with the workspace is becuase of the volume itself (because it needs to create a new PV and attach to the instance), so if instead I use emptyDir (which uses instance storage) maybe this would help. My issue is that I need to mock curl to read the post data, and this happens in two different containers (they are different steps) that was the way I found to save the post data. I accept ideas in the other PR on how to address that. I will try to use emptyDir instead if you agree.

johnbieren:
> > I don't think we should add a workspace to the task just to help unit tests. Adding a workspace is expensive (we already have tickets open to improve performance due to workspaces). Martin and I actually talked about it in the last week here [konflux-ci/release-service-catalog#761 (comment)](https://github.com/konflux-ci/release-service-catalog/pull/761#discussion_r1913110859)
> 
> If I understood correctly the issue with the workspace is due to the volume mapping itself (because it needs to create a new PV and attach to the instance), so if instead I use emptyDir (which uses instance storage) maybe this would help. My issue is that I need to mock curl to read the post data, and this happens in two different containers (they are different steps) and that was the way I found to save the post data. I accept ideas in the other PR on how to address that. I will try to use emptyDir instead if you agree.

I would say just change the mocked curl to echo some string then return 0 or 1. Without the workspace, you can't do things that we often do like in a check-result task, ensure that the mocked command was called the proper number of times and with the right arguments. I think that is okay... we have already gone without it in the other internal tasks that were migrated. This also limits what you can test. That is okay too. I think it is best to not have the workspace and just test what you can without it

theflockers:
the tests will be changed to not require worspace.

johnbieren:
> the tests will be changed to not require worspace.

thank you!

mmalina:
> > I don't think we should add a workspace to the task just to help unit tests. Adding a workspace is expensive (we already have tickets open to improve performance due to workspaces). Martin and I actually talked about it in the last week here [konflux-ci/release-service-catalog#761 (comment)](https://github.com/konflux-ci/release-service-catalog/pull/761#discussion_r1913110859)
> 
> If I understood correctly the issue with the workspace is due to the volume mapping itself (because it needs to create a new PV and attach to the instance), so if instead I use emptyDir (which uses instance storage) maybe this would help. 

It's not only requiring extra resources. It's also the fact that typically the whole pipelinerun will share just one PV. And if you use a workspace, this PV will be used. By default it's mounted in rw mode. This means that taskruns need to wait for one another, because two running taskruns cannot mount a PV in rw at the same time. We will try to use ro in some cases in the future, but with ro you wouldn't be able to use it for testing either.

emptyDir is fine, but it won't help you here either - we would typically have a check-result task to check the mock calls, but that's a separate task, so you won't have access to the emptyDir - the scope of that is limited to a single task.

PR #271, opened 2025-01-15
chore(RELEASE-1242): make add-fbc-contribution idempotent

theflockers:
This PR modifies the task add-fbc-fragment-to-index-image idempotent by checking IIB endpoint for builds for the current fbc_fragment, considering only prod builds and builds complete or in progress.

In addition, it contains code for RELEASE-1338 which the referenced ticket in the title depended on.

### DIFF
--- a/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
+++ b/internal-services/catalog/iib-add-fbc-fragment-to-index-image-task.yaml
@@ -4,7 +4,7 @@ kind: Task
 metadata:
   name: t-add-fbc-fragment-to-index-image
   labels:
-    app.kubernetes.io/version: "0.3.1"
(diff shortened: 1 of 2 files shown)


mmalina:
> This is to stable. Was the same change already merged in main?

Yeah, I have the same question.

theflockers:
This code was not promoted to stable yet as the overall task was finished just before the EOY freeze ( RELEASE-1242 and RELEASE-1338) and as I was mostly off the last week this code ended up not promoted.

We can ignore this one and wait this to be in production when the task is moved to the catalog, so it will include tests for this cases.

johnbieren:
@theflockers can we close this? The resources were long ago migrated to the catalog repo

mmalina:
Coincidentally, I noticed this today too :)

PR #272, opened 2025-01-15
fix: fbc release not resuming when in progress

theflockers:
this PR fixes the case when a in progress IIB build is to be resumed, it actually resumes it.


johnbieren:
You already have a PR moving this task to release-service-catalog, and as soon as it is merged, the task is being removed from this repo. Why open PRs to this repo for this task?

PR #273, opened 2025-01-20
chore(RELEASE-1327): bump staging manager images

johnbieren:
The new manager has support for InternalRequest.Spec.Pipeline so we can use resolvers.


johnbieren:
This will be done by the final pipeline once the RP change is merged

mmalina (approved):

PR #284, opened 2025-02-13
feat: removing iib pipeline (main)

theflockers:
this commit removes the IIB pipeline and tasks from the repository.


johnbieren (approved):

johnbieren:
I think yamllint, checkton and tknparse from here https://github.com/hacbs-release/app-interface-deployments/blob/main/.github/workflows/lint.yaml can be removed with this pr too

theflockers:
> I think yamllint, checkton and tknparse from here https://github.com/hacbs-release/app-interface-deployments/blob/main/.github/workflows/lint.yaml can be removed with this pr too

indeed. Was just looking at it.

johnbieren (approved):

PR #296, opened 2025-04-02
feat(RELEASE-1506): add new deployment

davidmogar:
This commit adds a new deployment for the new internal-services prod rh02 cluster.

### DIFF
--- a/internal-services/manager/production/manager-prod-rh02.yaml
+++ b/internal-services/manager/production/manager-prod-rh02.yaml
@@ -1,0 +1,106 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager-prod-rh02
+  namespace: stonesoup-int-srvc
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: deployment
+    app.kubernetes.io/instance: controller-manager-prod-rh02
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: internal-services
+    app.kubernetes.io/part-of: internal-services
+    app.kubernetes.io/managed-by: kustomize
+spec:
+  selector:
+    matchLabels:
+      control-plane: controller-manager
+  replicas: 1
+  template:
+    metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: manager
+      labels:
+        control-plane: controller-manager
+    spec:
+      # TODO(user): Uncomment the following code to configure the nodeAffinity expression
+      # according to the platforms which are supported by your solution. 
+      # It is considered best practice to support multiple architectures. You can
+      # build your manager image using the makefile target docker-buildx.
+      # affinity:
+      #   nodeAffinity:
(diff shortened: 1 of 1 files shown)


internal-services/manager/production/manager-prod-rh02.yaml
```diff
@@ -0,0 +1,106 @@
...
+          volumeMounts:
+            - mountPath: /mnt/internal-services/remote-client-config
+              name: remote-client-config
+              readOnly: true
+      serviceAccountName: controller-manager
+      terminationGracePeriodSeconds: 10
+      volumes:
+        - name: remote-client-config
+          secret:
+            secretName: remote-client-config-prod-rh02
```
mmalina, line 106: I think this will likely fail if the secret doesn't exist, so it should be added first, no? Not sure where though. Could it be that it's just done manually on the cluster? Because I don't see remote-client-config-prod-p02 defined anywhere either.

internal-services/manager/production/manager-prod-rh02.yaml
```diff
@@ -0,0 +1,106 @@
...
+            initialDelaySeconds: 5
+            periodSeconds: 10
+          # TODO(user): Configure the resources accordingly based on the project requirements.
+          # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+          resources:
+            limits:
+              cpu: 500m
+              memory: 512Mi
+            requests:
+              cpu: 10m
```
mmalina, line 90: I noticed that manager-prod-p02.yaml requests 100m. But manager-prod-rh01.yaml has 10m (I assumed you copied that one). Maybe safer to use 100m here?

internal-services/manager/production/manager-prod-rh02.yaml
davidmogar: Yes, it has to be defined in Vault. That's why I set it as a draft.

internal-services/manager/production/manager-prod-rh02.yaml
davidmogar: Sounds good to me. Updated.

internal-services/manager/production/manager-prod-rh02.yaml
mmalina: Sure, in Vault, but having it in Vault doesn't automatically mean it gets deployed. Maybe there is a wildcard somewhere to deploy all such secrets? But I don't see it anywhere.

davidmogar: You map it in app-interface: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/137976

mmalina: Sorry, somehow I thought this was about app-infra and Konflux deployments :) Makes sense.

mmalina (approved):

happybhati (approved):

PR #300, opened 2025-04-22
feat: adds rhoai quay.io pull-secret

theflockers:


johnbieren:
I don't understand why we would want this PR. Can you fill in the commit message with what this is for?

theflockers:
> I don't understand why we would want this PR. Can you fill in the commit message with what this is for?

my bad. Not sure why I did not add the ticket info and the msg.

mmalina (dismissed):

johnbieren:
I still don't think the commit message explains this change. I had to go to the ticket. The ticket talks about enabling the fbc opt in check for the openshift ai team. The commit talks about running internal requests. I also don't see any explanation in the commit (or ticket) why you are adding two secrets, not just one

theflockers:
> I still don't think the commit message explains this change. I had to go to the ticket. The ticket talks about enabling the fbc opt in check for the openshift ai team. The commit talks about running internal requests. I also don't see any explanation in the commit (or ticket) why you are adding two secrets, not just one

I understand @johnbieren, anyhow the ticket created does not reflect what should be done. In one of my comments in the ticket, I exposed that in conversation with @scoheb, we could not just link the secret to the generic `pipeline` account, as to do that, the secret would need to be created in all namespaces (due to the way appsre setups the shared secrets for the pipeline sa.
That said, the way to go would be adding the secret to the release-service-account, then linking the secret and running the internalrequest for FBC using it.

I will copy and paste what I just wrote in the ticket, but I disagre that explaining all the feature is necessary in the commit, besides what it is actually changing.

johnbieren:
> I disagre that explaining all the feature is necessary in the commit, besides what it is actually changing

Well, I didn't say that. I said I don't understand why we want this. Then, I want to the ticket and got some context but that it still doesn't make sense. I don't see anything in the commit message about why you are adding two secrets instead of one. Your commit message says this
```
this PR links the rhoai quay.io pull-secret to the
release-service-account, which should be used to
run the internal-requests.
```
yet you link two secrets. So, no, we certainly do not need to describe entire features in commit messages, but the commit message should describe what you are doing and why.

theflockers:
Please take a look @johnbieren

johnbieren (approved):

mmalina:
I also don't understand. @theflockers you're saying that in order to enable the new secret, all internal requests will now need to run as the SA instead of "pipeline"? Where was that changed? And that's why two secrets are being added - the quay.io one was previously accessible to the pipeline SA, but now we need it for this other SA? What does it actually give access to?

theflockers:
> See also [konflux-ci/release-service-catalog#915](https://github.com/konflux-ci/release-service-catalog/pull/915)
 
 I left the PR link in the commit message, and also updated the description.

PR #309, opened 2025-06-05
fix: script should include only finished plr

theflockers:
the script should include only finished pipelineruns


theflockers:
> The commit title says only running ones and the commit body says only finished ones

fixed. Finished is the correct.

johnbieren (approved):
So before it just checked if completionTime exists, but now it makes sure it is set to some time?

PR #317, opened 2025-07-23
feat(RELEASE-1768): add rh03 manager

davidmogar:
This commit add a new deployment for the production cluster rh03.


mmalina:
Note for future reference: This is the same as https://github.com/hacbs-release/app-interface-deployments/pull/315 from yesterday, but for production (the stable branch)

PR #326, opened 2025-09-01
chore(RELEASE-1828): convert utils image to digest format

seanconroy2021:
Switch the utils image in the CronJob from a tag to a  digest so the promotion workflow can run. Image digest is from main.


mmalina (approved):

johnbieren:
Was this done in main too?

seanconroy2021:
Yeah its already changed in main to a digest

PR #347, opened 2025-09-25
chore(RELEASE-1863): scale prod p01 manager to 0 replicas

johnbieren:
This commit scales the stone-prod-p01 manager replicas to 0. This will turn off the app-sre cluster for stone-prod-p01, which will be replaced with the common cluster.


johnbieren:
Ready for review, but not to be merged

mmalina (approved):

